PR #38, opened 2014-12-04
djinni doesn't respect namespace'd cpp objects always

skabbes:
Applies a more incremental fix for https://github.com/dropbox/djinni/pull/28


skabbes:
Actual fix was landed here: https://github.com/dropbox/djinni/commit/21fecda45fb9a553ade6c9d95ee0e52f9f83d733

PR #40, opened 2014-12-12
Added option --objc-base-lib-include-prefix

mknejp:
Extracted from #28 and based on the latest master for easier review


j4cbo:
Pushed as 86e6e0d39cf4723df25aff135f205fab87ddbdff

PR #47, opened 2015-01-14
Generate interfaces in Java when possible

mrdomino:
If the djinni interface doesn't contain constants or static members, then it
can be represented as an interface (not abstract class) in Java.

### DIFF
--- a/example/generated-src/java/com/dropbox/textsort/TextboxListener.java
+++ b/example/generated-src/java/com/dropbox/textsort/TextboxListener.java
@@ -3,6 +3,6 @@
 
 package com.dropbox.textsort;
 
-public abstract class TextboxListener {
-    public abstract void update(ItemList items);
+public interface TextboxListener {
+    public void update(ItemList items);
 }
--- a/example/handwritten-src/java/com/dropbox/textsort/TextboxListenerImpl.java
+++ b/example/handwritten-src/java/com/dropbox/textsort/TextboxListenerImpl.java
@@ -4,7 +4,7 @@ import android.widget.EditText;
 
 import java.util.ArrayList;
 
-public class TextboxListenerImpl extends TextboxListener {
+public class TextboxListenerImpl implements TextboxListener {
 
     private EditText mTextArea;
 
--- a/src/source/JavaGenerator.scala
+++ b/src/source/JavaGenerator.scala
@@ -143,7 +143,10 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
       writeDoc(w, doc)
 
       javaAnnotationHeader.foreach(w.wl)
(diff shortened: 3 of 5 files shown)


mrdomino:
Any feedback on this? It makes it a lot easier to use Djinni-generated classes in Android apps, e.g. by implementing a listener interface while still extending Activity to get easy access to runOnUiThread.

mrdomino:
Aha. It seems that this change may violate some assumptions of the JNI support code. When we implement multiple djinni interfaces at once, our app crashes with jniExceptionCheck, and we see "JNI WARNING: DeleteLocalRef(0x421d2948) failed to find entry" right beforehand.

Is this a known issue?

Bennyyyy:
+1

mrdomino:
Bump.

mknejp:
I do agree that `+j`interfaces should create an `interface`. Constants aren't a problem (an `interface` can have `public static final` member values), however the Objective-C generator does support static methods (though I haven't see a use for it yet and it won't be allowed for `+j +o`). However, making the creation of `class` versus `interface` conditional just means a lot of code can break simply because someone added a static method somewhere in a Djinni file.

> Aha. It seems that this change may violate some assumptions of the JNI support code. When we implement multiple djinni interfaces at once, our app crashes with jniExceptionCheck, and we see "JNI WARNING: DeleteLocalRef(0x421d2948) failed to find entry" right beforehand.

It would help to recreate your situation in a test as a starting point, even if you can't find a fix.

mrdomino:
I actually am having a hard time reproducing that issue.

artwyman:
This is clever, but I share @mknejp's concerns about the automatic nature of it causing easy build breakages.  In its current state, I think the feature should be optional, controlled by a command-line switch.

There might be a better alternative if we can raise the requirement to Java 8, which will be supported by Android officially soon.  That will allow static methods on interfaces.  That just leaves the static inner class for the auto-generated CppProxy as a blocking factor, but I think that could be moved out into a separate Java class (package-private) without it being a big deal.  I think something like that, which would allow all Djinni interfaces to be Java interfaces would be a better feature in the long term.

itsff:
@artwyman For the time being, could we at least generate Java interfaces for cases where no static methods are present? As you suggested, the CppProxy class could be implemented (via convention) as a separate package private class. This would really help for cases where a Java class is supposed to implement many listeners for callbacks.

artwyman:
@itsff that's what this original PR proposed to do, and I think it's a reasonable feature.  The open issues are making it optional behavior (so that existing code doesn't break unnecessarily because someone adds/removes a static method), and figuring out if the proxy cache really has problems with this scenario (it's designed not to, since the same can come up in Objective-C with protocols, but it hasn't been tested).

artwyman:
Supplanted by #288

PR #50, opened 2015-01-23
Add sample file and documentation for NDK path specification

mbrgm:
When trying to build the android example, some users might run into the problem
where the path to the Android NDK is not properly setup, so gradle fails to
build the project.

As proposed in #8, this PR adds a `local.properties.sample` file, as well as a
troubleshooting section in the README.md, which describes how to solve the
problem.


j4cbo:
Thanks!

PR #57, opened 2015-02-06
Don't apply cppHeaderExt to JNI base libs

mrdomino:
Otherwise, JNI support lib includes break if hpp-ext is set to anything other than hpp.


marcomagdy:
+1 for this request

meoguru:
+1, fixed with #65.

j4cbo:
Thanks!

PR #60, opened 2015-02-09
Support calls into JNI on a new thread by attaching to it.

theojulienne:
Also added some output for this case so it’s more obvious what went wrong, rather than just std::aborting.

### DIFF
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -18,6 +18,8 @@
 #include <cassert>
 #include <cstdlib>
 
+#include <iostream>
+
 static_assert(sizeof(jlong) >= sizeof(void*), "must be able to fit a void* into a jlong");
 
 namespace djinni {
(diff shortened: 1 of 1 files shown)


canatella:
Mmm, don't you need to detach the thread at some point ?

theojulienne:
Ideally, and if threads are being created and destroyed, then yes. I didn't implement this as a separate attach/detach call because djinni doesn't really expose a public API to C++ code to perform actions like this, instead it sort of just magically makes things work.

The specific case where this happened was when a native C/C++ library has an internal thread that sends callbacks/events out on that thread. Because we needed to get the events handled on the UI thread, we used an event_loop->post (similar to that of mx3) to get it there, but of course it crashes out because the calling thread wasn't started through the JVM, and the push to UI thread is implemented in Java. ObjC didn't seem to have this issue, it doesn't seem to need additional context on thread calls, or it handles adding and removing it internally.

Another issue to consider is that sometimes you may not have the facility to inject something, like a thread attach, to the start-up of a thread from a library without modifying said library.

Looking at the code again, I probably shouldn't have left the ```std::abort``` in there at all, since exceptions would work fine.

skabbes:
I think this exact issue has been discussed with djinni, and ended up deciding that it was a non-goal of the project.  There are several issues with C++ native threads that interact with the JNI / Java, one being the thread needs to attach to the JNI.  Although [the android docs](http://developer.android.com/training/articles/perf-jni.html#faq_FindClass) advise against launching threads from C++ (something about the class loader not being correct).  So I think this is an issue that djinni doesn't (and shouldn't) solve - at least in the fashion.

Check out [this discussion](https://github.com/libmx3/mx3/issues/37) on mx3 for some more context.  And the current solution in mx3 is to have an `thread_launcher = interface +o +j` interface for each platform.

hamedmohammadi:
guys you are saying that we should not directly use jni from c++ threads . but we can use it in normal usage  .am i right?

canatella:
You can but you need to attach/detach the thread by yourself.

2015-03-09 7:00 GMT+01:00 hamed.nova <notifications@github.com>:

> guys you are saying that we should not directly use jni from c++ threads .
> but we can use it in normal usage .am i right?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dropbox/djinni/pull/60#issuecomment-77804366>.
>



-- 
Damien MERENNE   <dam@cosinux.org>

PR #61, opened 2015-02-11
Avoid unnecessary conversion and temporaries

marcomagdy:
Passing the result of c_str() to a function that takes std::string as argument is slow and redundant.


j4cbo:
Good catch, thanks! Can you fill in a CLA form at https://opensource.dropbox.com/cla/ and then I'll merge this?

marcomagdy:
Done.
Thanks!

PR #65, opened 2015-03-07
Fix incorrect extensions of support header files

### LINKED ISSUE #64: Incorrect extensions of Djinni support header files
I called Djinni with option `--hpp-ext h` but the extensions of Djinni support header files are also changed too, for example there is `HString.h` in a generated file instead of `HString.hpp`.

meoguru:
This pull request is to fix #64, please review.


meoguru:
FYI: Already signed the CLA.

meoguru:
Closed due to duplicate with #57. Thanks.

Issue #70, opened 2015-04-08
A Barrel of Refactoring


mknejp:
Hey folks,

over the past month or two Djinni was used quite extensively here and we ran into a few problems, resulting in a number of refactorings, both internal (not changing the output) and external (significantly changing the output). Besides, we need something similar to #52, which at the time posed quite a challenge considering how Djinni was structured. So in preparation to make #52 (and others like #54, #46, #45) much easier to realize something had to happen. 

Before doing a pull request I thought it'd be better to share these changes and get your input. So here we go:

* Split the `ObjcGenerator` into `ObjcGenerator` and `ObjcppGenerator`. This makes it easier to adjust only one of them and is also how the Java/JNI generators are separated. The philosophy here is that using only `ObjcGenerator`is sufficient to create all the Obj-C files necessary in the same fashion as `JavaGenerator` only creates Java classes, but no JNI content.
* Introduce "Marshal" types, responsible for naming types, files, creating parameter signatures, to/from C++ conversion, etc. Now every language has its own Marshal. The `xxxGenerators` are responsible for the overall file structure, whereas `xxxMarshals` do the actual language interop. So now `JNIGenerator` can make use of `CppMarshal` to create consistent type signatures etc. (This is a crucial part to make #52 remotely possible)
* The Obj-C support library was restructured similar to how the JNI support library works, using template translation types. This *wastly* simplifies the work of ObjcppGenerator.
* Before, Djinni was generating `@protocol` types for both `+c` and `+o` and provided a `xxCppProxy` in a `+Private.h` include file. This was a real thorn to every Obj-C developer exposed to Djinni generated types as it didn't really follow any existing conventions on that platform. Therefore came an implementation of #25. To recap: `+c` types now generate an `@interface` acting as proxy to C++ and `+o` types generate a `@protocol` to be implemented in Objc-C. Therefore `+c` types are represented by `T*` variables and `+o` by `id<T>`. There was similar outcry in the Java camp about `+j` not creating a Java `interface` instead of `class` but changing the JNI side is a bit more involved.
* Since Obj-C and Obj-C++ are now considered two separate generators, there's a new set of `--objcpp-x` command line arguments:
  * `--objcpp-out`: The output folder for private Objective-C++ files (Generator disabled if unspecified).
  * `--objcpp-include-prefix`: The prefix for `#import` of Objective-C++ header files from Objective-C++ files.
[...]

PR #39, opened 2014-12-12
default namespace fixed to "djinni_generated"

mknejp:
Extracted from #28 and based on the latest master for easier review

PR #74, opened 2015-04-15
Feature/codegen refactor

mknejp:
More details about this can be found here: #70 

This includes #40 and #39.


example/android/app/jni/Application.mk
```diff
@@ -4,8 +4,8 @@ APP_ABI := all
 APP_OPTIM := release
 APP_PLATFORM := android-8
 # GCC 4.8 Toolchain - requires NDK r9
-NDK_TOOLCHAIN_VERSION = 4.8
+NDK_TOOLCHAIN_VERSION = clang
```
skabbes: Any particular reason for the lib and compiler changes?

src/source/CppGenerator.scala
```diff
@@ -180,56 +147,52 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
...
+          w.wl(marshal.fieldType(f.ty) + " " + idCpp.field(f.ident) + ";")
         }
 
-        w.wl
         if (r.derivingTypes.contains(DerivingType.Eq)) {
-          w.wl(s"bool operator==(const $actualSelf & other) const;")
-          w.wl(s"bool operator!=(const $actualSelf & other) const;")
+          w.wl
+          w.wl(s"friend bool operator==(const $actualSelf& lhs, const $actualSelf& rhs);")
+          w.wl(s"friend bool operator!=(const $actualSelf& lhs, const $actualSelf& rhs);")
```
skabbes: +1

example/android/app/jni/Application.mk
mknejp: I'm actually not sure anymore. I think I was trying it out on both compilers for warnings and probably forgot to discard the change. Maybe it was because of the need for C++14 in the support lib. I don't see a reason why it shouldn't work with GCC

izacus: Last I tested the 4.8 GCC toolchain was missing `std::experimental:optional` which prompted the switch in our code.

j4cbo:
So I'm slowly cherry-picking the individual changes here to go through the necessary updates to our code that uses it. It looks like the JNI codegen change has issues compiling the test suite:

     [exec] In file included from ../generated-src/jni/NativeNestedCollection.cpp:5:
     [exec] ../../support-lib/jni/Marshal.hpp:245:32: error: no member named 'Boxed' in 'djinni::Set<djinni::String>'
     [exec]                 using EJniType = typename T::Boxed::JniType;
     [exec]                                           ~~~^
     [exec] ../generated-src/jni/NativeNestedCollection.cpp:17:70: note: in instantiation of template class 'djinni::List<djinni::Set<djinni::String> >' requested here
     [exec]                                                            ::djinni::List<::djinni::Set<::djinni::String>>::fromCpp(jniEnv, c.set_list).get())};
     [exec]                                                                      ^
     [exec] In file included from ../generated-src/jni/NativeMapListRecord.cpp:5:
     [exec] ../../support-lib/jni/Marshal.hpp:245:32: error: no member named 'Boxed' in 'djinni::Map<djinni::String, djinni::I64>'
     [exec]                 using EJniType = typename T::Boxed::JniType;
     [exec]                                           ~~~^
     [exec] ../generated-src/jni/NativeMapListRecord.cpp:17:70: note: in instantiation of template class 'djinni::List<djinni::Map<djinni::String, djinni::I64> >' requested here
     [exec]                                                            ::djinni::List<::djinni::Map<::djinni::String, ::djinni::I64>>::fromCpp(jniEnv, c.map_list).get())};
     [exec]                                                                      ^

I get the same errors on the latest `feature/codegen-refactor` in your fork.

Can you rebase the remaining changes on top of current master and fix the compilation issues? Thanks!

mknejp:
Hmm it seems my test setup was borked and the Java tests weren't running properly, weird.
Not every single commit may compile and run tests on its own though, I have to admit that I often do several commits in sequence and then regenerate/run tests before pushing the commits to keep the diffs focused. Not very helpful in this situation.

Anyway, since trying to rebase the existing branch is now giving a bazillion conflicts I'll have to make a new one and try to split it over separate pull requests where I can. Sorry for the inconvenience.

Issue #76, opened 2015-04-29
JNI Exception check in HBinary.hpp calls forbidden JNI methods in critical section


izacus:
When retrieving `binary` data from Java environment, `fromJava` method in [HBinary.hpp](https://github.com/dropbox/djinni/blob/b5bacd702e878c2a1cb0b474410190d82207c135/support-lib/jni/HBinary.hpp) calls `jniExceptionCheck` which calls further methods.

The exception check call is made between the time `GetPrimitiveArrayCritical` and `ReleasePrimitiveArrayCritical` is called, causing undefined behaviour forbidden by the [JVM standard](https://docs.oracle.com/javase/6/docs/technotes/guides/jni/spec/functions.html).

It also causes a hard crash on Android ART / CheckJNI with an exception:

```
JNI DETECTED ERROR IN APPLICATION: thread Thread[13,tid=18055,Runnable,Thread*=0xaf075000,peer=<redacted>] using JNI after critical get
```

Suggest removing the check call in that method since it's usually done by the callee anyway.

PR #77, opened 2015-04-29
Fixed illegal exception check in HBinary.hpp

izacus:
This is a proposed fix for issue #76 - JNI exception check is done before and after holding a critical reference to a primitive array.


j4cbo:
Good catch, thanks. I've merged in a fix using a slightly different strategy in 9996b44195f20feba24aa27d97b244e749f37bf9

PR #84, opened 2015-05-14
Freeze gyp at the last version that supports Android.

marcinkaszynski:
Gyp no longer supports Android, this commit freezes it at the last version that did.


marcinkaszynski:
I created this PR based on master by mistake.  I'm closing it, will create another one from a branch.

PR #88, opened 2015-05-19
Move declaration of the extended class for the extended record into the namespace.

greenhat:
When you define an extended `record +c` declaration of the extended class was placed outside the namespace. In our case `Rect` class was declared outside the namespace and it was in conflict with existing struct `Rect` (from MacTypes.h) already being imported in the global namespace.


greenhat:
Fixed in other PR. Thanks.

PR #92, opened 2015-05-26
Enable NRVO in JNI's Binary::toCpp

mknejp:
This optimization is usually disabled if not all paths return the same local object.


j4cbo:
Good catch, thanks!

PR #94, opened 2015-05-26
Java nullity annotations

FuegoFro:
Add the ability to specify Java nullity annotations which are added to constants, return values, and parameters based on whether or not they're optional.


j4cbo:
Thanks! Can you update this to also include the changes to the generated source? (Run `run_djinni.sh` in `support-lib/` and `example/`, then amend the commit with the result)

FuegoFro:
I didn't actually add nullity annotations for the test or example sources, which is to say I didn't add the `java-nullable-annotation` or `java-nonnull-annotation` arguments in the call to djinni in those scripts. Would you like me to do so? If so, do you care which nullity annotations are used?

The ones we use regularly are `javax.annotation.CheckForNull` and `javax.annotation.Nonnull` but there are [a few others](http://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use). In any case, we'd have to add a library to the test suite such that these annotations are defined at compile time.

FuegoFro:
I added the annotations, using the two that I mentioned :)

src/source/JavaGenerator.scala
```diff
@@ -89,6 +92,7 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
     for (c <- consts) {
       writeDoc(w, c.doc)
       javaAnnotationHeader.foreach(w.wl)
+      marshal.nullityAnnotation(c.ty).foreach(w.wl)
```
j4cbo: Hm, I don't think this makes sense - this annotation is on the class itself, where neither nullable or checkfornull would make sense, right?

FuegoFro: This is on a constant field, not a class definition (or I'm very confused here). In fact, looking at this again, it kinda seems like maybe the previous line (the `javaAnnotationHeader.foreach(w.wl)` line) shouldn't actually be here?

j4cbo: Doh, my bad.

Issue #52, opened 2015-01-29
Preexisting record types


skabbes:
Djinni really only needs to know how to convert record types between objc, java and c++.  What if we extended the include format to take plugins like this?

@include "date.json"

```coffeescript
djinni_type: "custom_date"
objc_type: "NSDate"
java_type: "java.util.Date"
cpp_include_path: "<chrono>"
cpp_type: "std::chrono::system_clock::time_point"
jni_type: "jobject"
```

```cpp
// custom_date.hpp
namespace djinni {
  jobject custom_date_to_java(time_point);
  NSDate * custom_date_to_objc(time_point);
  time_point custom_date_to_cpp(jobject);
  time_point custom_date_to_cpp(NSDate *);
}
```

mknejp:
This would certainly help making stuff more modular. We are currently struggling with how to setup Djinni in na environment with libraries. Say library A lives in its own repository and has Djinni-generated classes, and library B uses library A and has its own djinni-generated classes that need to interact with A in their interface, then there is no *easy* way to make this work without exposing A's djinni files and hope for no linker errors or ODR violations.

So bonus points if Djinni was able to generate these JSON files for inclusion of generated interfaces in other projects.

pwais:
djinni currently deep-copies all record types.  It would be nice if this proposed feature didn't *necessarily* require deep copies.  E.g. a user could leverage it to effectively share state across the language boundary (much like interfaces do currently).  Shouldn't be challenging to support if the user-provided type mapping function can have arbitrary behavior.

+1 as using this proposed feature to solve #54 !

mknejp:
So, I was on a coding spree and this is now a thing. The thing can be found here: https://github.com/mknejp/djinni/tree/feature/extern-types and is based on the refactoring done in #54. It was actually pretty easy to add once all the necessary machinery was in place.

This contains two main features:
* The ability for Djinni to dump information about the types its generating into a config file and
* the ability to read such files back in and use the stored information for code generation.

I decided to go with YAML instead of JSON for these main reasons:
* YAML can store comments.
* YAML can store multiple documents in a single file, so you can have a single `mylib.yaml` file containing all the types in one place.
* JSON is a subset of YAML so it can still be read if really needed.
* *Much* more human-friendly format

YAML generation is controlled by the following options:
* `--yaml-out`: The output folder for YAML files (Generator disabled if unspecified).
[...]

PR #95, opened 2015-05-27
External types

mknejp:
This is the implementation foreshadowed in #52 based on the latest and greatest master.

Long story short: This allows Djinni to generate YAML files for storing all the information necessary to integrate the generated types in a different project. The directive `@extern "mylib.djinni"` makes the types within available to your Djinni project.

These YAML files can also be created by hand and, provided with the necessary translators, allow the integration of types not generated by Djinni (e.g. `std::chrono::duration`<-> `NSTimeInterval` <-> `java.time.Duration`).

I added a section to the README and there are some examples in the test suite. One of them mimics the builtin "date" type, showing that complexity can be removed from the generators for types that don't require special magic. However that leaves the open question of how to inject these "standard" types. The other example is `duration`:
```
@extern "duration.yaml"

r = record {
  h: duration<i32, h>;
  min: duration<i32, min>;
  ns: duration<i32, ns>;
}
```
Something to keep in mind is the YAML file stores all the include paths as they were specified in the Djinni invocation. For some types that includes paths to the support lib, so maybe it would be wise to make a standard `include` directory for the "standard" headers and require people to set their search paths to it.

This feature does not change the output of existing stuff.


mknejp:
rebased

zaren678:
+1

mknejp:
rebased

PR #104, opened 2015-06-19
Allow imported IDL files to be in same directory as parent IDL file

jrogers:
Java.io.File.getParent() previously would return null and the import file path will end up as "null/file.djinni", which fails.
Use case is you have a bunch of IDL files in a directory and have a single file that imports them all, ie.

main.djinni \
    @import "filename.djinni"

filename.djinni \
    ...


jrogers:
Needs an update

PR #112, opened 2015-07-08
Adding more Android error troubleshooting to example readme

spanndemic:
I was working through the example project today, and ran into some additional Android error messages.

I've added the errors I encountered and the solutions to the example's readme, so hopefully they will help others getting started with Djinni.

A couple of the issues may be due to different versions of Gradle, and changing a line or two in the .gradle file seems  to fix them. I was worried that this would break the example for users with different versions of Gradle, however, so didn't implement the fix in the source.

Great job and many thanks for such a useful tool!


j4cbo:
Awesome, thanks! Can you fill in our CLA at https://opensource.dropbox.com/cla/ and then I'll merge this?

spanndemic:
done!

PR #117, opened 2015-07-14
Generate convenience initializers for Objective-C records

mknejp:
Generate convenience initializers for Objective-C records

This provides a consistent pair of initializers for all records:
`- (nonnull instancetype)initWithArg:(A*)arg;`
`+ (nonnull instancetype)myRecordWithArg:(A*)arg;`

The static convenience initializer is only created for records *without* the +o extension. This also applies to records with no members. Now even empty records have a designated initializer.


j4cbo:
Oh, sweet. I hadn't seen `instancetype` before.

PR #122, opened 2015-07-29
Add a makefile rule to _only_ build djinni proper

DEGoodmanWilson:
Unclear if this is the best way to do it, but I wanted a way to build djinni from the top-level directory without having to build any examples or tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,4 +43,4 @@ example_android: GypAndroid.mk
 test:
 	make -C test-suite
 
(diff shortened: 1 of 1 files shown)


j4cbo:
Looks good. You'll also need to add `djinni` to the `.PHONY` line at the end (else the rule will break if a file called djinni ever appears).

DEGoodmanWilson:
Done!

PR #123, opened 2015-07-31
Returning immutable Objective-C collections when converting from C++

NachoSoto:
The specified `ObjcType` in all of these is the immutable counterpart (`NSArray,` `NSSet`, and `NSDictionary`) so returning a mutable instance was unnecessary. This has a few added benefits:
- **Guarantees immutability**: even though the types were immutable, one can (though obviously discouraged) mutate the instances.
- **Optimizes memory usage**: it's common in Objective-C to perform *defensive copying* (I've seen Djinni does this in a few places). By copying early we guarantee that only one copy is made, rather than each class creating its own copy (remember that calling `-copy` on an immutable instance simply returns `this`).
- **Optimizes performance**: immutable implementations of these classes in `Foundation` are potentially better optimized for searching and fetching, since they don't need to be optimized for insertion, deletion, or update.

### DIFF
--- a/support-lib/objc/DJIMarshal+Private.h
+++ b/support-lib/objc/DJIMarshal+Private.h
@@ -203,7 +203,7 @@ public:
         for(const auto& value : v) {
             [array addObject:T::Boxed::fromCpp(value)];
         }
-        return array;
+        return [array copy];
     }
 };
 
@@ -233,7 +233,7 @@ public:
         for(const auto& value : s) {
             [set addObject:T::Boxed::fromCpp(value)];
         }
-        return set;
+        return [set copy];
     }
 };
 
@@ -266,7 +266,7 @@ public:
         for(const auto& kvp : m) {
             [map setObject:Value::Boxed::fromCpp(kvp.second) forKey:Key::Boxed::fromCpp(kvp.first)];
         }
-        return map;
+        return [map copy];
     }
 };
 


NachoSoto:
Gentle bump :)

NachoSoto:
Bump?

artwyman:
Hmm, I'm not sure what standard practice is for ObjC libraries.  While this approach seems safer, I dislike the fact that it ensures an extra copy (albeit a shallow one) on each marshal, as opposed to the theoretical ability to save a copy if more than one user defensively copies the result.  Since we always create new objects across the boundary, Djinni never causes sharing of records and collections, though the user can do whatever they want.  I kinda wish there were a "best of all worlds" alternative to simply mark a mutable collection as immutable, letting it change representation internally if it thinks it should.

@j4cbo any thoughts?

artwyman:
@NachoSoto can you re-run the generators for the example and test-suite code so we can see the results of this on the generated code?

j4cbo:
I think we should do this. IIRC we actually did copy into immutable types for a while, but it got lost in a refactor.

NachoSoto:
> I kinda wish there were a "best of all worlds" alternative to simply mark a mutable collection as immutable

There is no way, as `NSArray

NachoSoto:
> can you re-run the generators for the example and test-suite code so we can see the results of this on the generated code

This doesn't actually change the generated code, only `DJIMarshal+Private.h`. #124 does change the generated code, and I ran the generator so that diff shows the result.

artwyman:
Also @NachoSoto can you sign the CLA before merge? https://opensource.dropbox.com/cla/

artwyman:
Sorry, going too much on auto-pilot.  This diff is only support-lib code, so no generated code is involved.

j4cbo:
There's a potential further optimization we could do here, by marshaling into a temporary raw C array (would likely have to be heap-allocated, since VLAs are hazardous, and don't exist in C++ anyway) and then using `[NSArray arrayWithObjects: count: ]` et al. But this is a good start.

j4cbo:
(I think this is ready for merge pending CLA)

NachoSoto:
Thanks!

PR #124, opened 2015-07-31
Objective-C generated classes copy collection types in constructor

NachoSoto:
This is a small follow up to #123. The `ObjcGenerator` was already correctly set up to automatically copy `NSString` instances to ensure that if an `NSMutableString` instance was passed instead, an immutable copy was created.
With this change the same is true for `NSSet`, `NSArray`, and `NSDictionary` properties.

Note that thanks to #123 this should not have a performance impact (neither memory or CPU), since at most only one copy will be made.

### DIFF
--- a/src/source/ObjcGenerator.scala
+++ b/src/source/ObjcGenerator.scala
@@ -187,6 +187,9 @@ class ObjcGenerator(spec: Spec) extends Generator(spec) {
     def checkMutable(tm: MExpr): Boolean = tm.base match {
       case MOptional => checkMutable(tm.args.head)
       case MString => true
+      case MList => true
+      case MSet => true
+      case MMap => true
       case MBinary => true
       case _ => false
     }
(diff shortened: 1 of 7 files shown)


NachoSoto:
Bump?

artwyman:
@NachoSoto can you re-run the generators for the example and test-suite code so we can see the results of this on the generated code?

See #123 for my thoughts on copying, which we should discuss a bit more before merging.

artwyman:
Also @NachoSoto can you sign the CLA before merge? https://opensource.dropbox.com/cla/

NachoSoto:
> can you re-run the generators for the example and test-suite code so we can see the results of this on the generated code?

I already did. You can see the examples, like `DBMapDateRecord.mm`

NachoSoto:
> Also @NachoSoto can you sign the CLA before merge? https://opensource.dropbox.com/cla/

Done

artwyman:
Sorry, going too much on auto-pilot.  I didn't distinguish which type of code (generated or hand-written) I was looking at.

NachoSoto:
Thanks!

PR #126, opened 2015-08-03
Android Makefile / UI changes

mmizzle9:
Please merge this by having djinni Makefile check for an NDK environment variable and then look for ndk.dir like it had been done with the sdk.dir in other places.


artwyman:
Sorry this has sat unanswered for so long.  It pre-dates my time as Djinni maintainer.

I can't merge this directly unless you sign the CLA, but I'll go ahead and make equivalent changes on my own, so I'll close this out.

I don't want to make the environment variable a requirement for Djinni and break existing users, but I can use a fallback mechanism which we use in our internal gyp scripts.  There also seems to be some disagreement online about whether NDK_HOME or ANDROID_NDK_HOME is the right variable to use.  I'll go with ANDROID_NDK_HOME, which is what we adopted from docs elsewhere.

PR #129, opened 2015-08-13
Returning immutable Objective-C collections when converting from C++ (alternative approach)

steipete:
This is an alternative approach for https://github.com/dropbox/djinni/pull/123 and https://github.com/dropbox/djinni/pull/124 from @NachoSoto, and I included one commit from him here.

Was playing around with various options, and in the end only using C made things significantly faster.
(VLA arrays are too risky since stack-based and vector only added a barely measurable benefit)

Here's a small, non-scientific benchmark on the iOS 9 Simulator:

```objc
    const size_t runs = 10, elements = 10000000;

    std::vector<NSString *> v;
    for (NSUInteger i = 0; i < elements; i++) {
        v.push_back([NSString stringWithFormat:@"element %tu", i]);
    }

    __block NSArray *oldResult, *newResult;

    uint64_t time_new = dispatch_benchmark(runs, ^{
        id __strong *const array = (id __strong *const)calloc(v.size(), sizeof(id));
        assert(array);
        NSUInteger idx = 0;
        for(const auto& value : v) {
            array[idx++] = value;
        }
        newResult = [NSArray arrayWithObjects:array count:v.size()];
        free(array);
    });
    NSLog(@"new: %f ms", time_new/1e6);

    uint64_t time_old = dispatch_benchmark(runs, ^{
        auto array = [NSMutableArray arrayWithCapacity:static_cast<NSUInteger>(v.size())];
        for(const auto& value : v) {
            [array addObject:value];
        }
        oldResult = [array copy];
    });
    NSLog(@"old: %f ms,", time_old/1e6);
    assert([oldResult isEqualToArray:newResult]);

    // Sprinkle in some CF, just out of curiosity.
    uint64_t time_cf = dispatch_benchmark(runs, ^{
        auto array = CFArrayCreateMutable(nullptr, static_cast<CFIndex>(v.size()), &kCFTypeArrayCallBacks);
        for(const auto& value : v) {
            CFArrayAppendValue(array, (void *)value);
        }
        oldResult = [NSArray arrayWithArray:(__bridge NSArray *)array];
        CFRelease(array);
    });
    NSLog(@"cf: %f ms,", time_cf/1e6);
```

Outputs on my machine:

```
2015-08-13 13:04:46.302 PSPDFCatalog[12832:366832] new: 522.163530 ms
2015-08-13 13:04:54.225 PSPDFCatalog[12832:366832] old: 769.401330 ms,
2015-08-13 13:05:04.047 PSPDFCatalog[12832:366832] cf: 955.982234 ms,
```

So it's about 30% faster, but we end up with a `calloc` in the code. Not sure if that's worth it.

### DIFF
--- a/support-lib/objc/DJIMarshal+Private.h
+++ b/support-lib/objc/DJIMarshal+Private.h
@@ -199,14 +199,13 @@ public:
 
     static ObjcType fromCpp(const CppType& v) {
         assert(v.size() <= std::numeric_limits<NSUInteger>::max());
-
         NSUInteger const count = static_cast<NSUInteger>(v.size());
-        NSUInteger idx = 0;
-        EObjcType __strong array[count];
+        std::vector<EObjcType> array;
+        array.reserve(count);
         for(const auto& value : v) {
-            array[idx++] = T::Boxed::fromCpp(value);
+            array.push_back(T::Boxed::fromCpp(value));
         }
-        return [NSArray arrayWithObjects:array count:count];
+        return [NSArray arrayWithObjects:array.data() count:count];
     }
 };
 
@@ -233,12 +232,12 @@ public:
     static ObjcType fromCpp(const CppType& s) {
         assert(s.size() <= std::numeric_limits<NSUInteger>::max());
         NSUInteger const count = static_cast<NSUInteger>(s.size());
-        NSUInteger idx = 0;
-        EObjcType __strong set[count];
+        std::vector<EObjcType> set;
+        set.reserve(count);
         for(const auto& value : s) {
-            set[idx++] = T::Boxed::fromCpp(value);
+            set.push_back(T::Boxed::fromCpp(value));
         }
-        return [NSSet setWithObjects:set count:count];;
+        return [NSSet setWithObjects:set.data() count:count];
     }
 };
 
@@ -268,14 +267,15 @@ public:
     static ObjcType fromCpp(const CppType& m) {
         assert(m.size() <= std::numeric_limits<NSUInteger>::max());
         NSUInteger const count = static_cast<NSUInteger>(m.size());
-        NSUInteger idx = 0;
-        ObjcKeyType __strong keys[count];
-        ObjcValueType __strong objects[count];
+        std::vector<ObjcKeyType> keys;
+        keys.reserve(count);
+        std::vector<ObjcValueType> objects;
+        objects.reserve(count);
         for(const auto& kvp : m) {
-            keys[idx] = Key::Boxed::fromCpp(kvp.first);
-            objects[idx++] = Value::Boxed::fromCpp(kvp.second);
+            keys.push_back(Key::Boxed::fromCpp(kvp.first));
+            objects.push_back(Value::Boxed::fromCpp(kvp.second));
         }
-        return [NSDictionary dictionaryWithObjects:objects forKeys:keys count:count];
+        return [NSDictionary dictionaryWithObjects:objects.data() forKeys:keys.data() count:count];
     }
 };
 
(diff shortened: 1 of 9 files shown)


steipete:
I don't think it's worth it, feel free to close and merge the above mentioned PRs instead. One interesting line is https://github.com/dropbox/djinni/pull/129/files#diff-61477e190b9e4951838edf1fa955dd19R74 though; which would bring the public/private ObjC pairs more inline with what people typically use (Where the private header always includes the public one, since it is an extension with added functionality)

support-lib/objc/DJIMarshal+Private.h
```diff
@@ -199,11 +199,17 @@ class List {
...
+        NSUInteger idx = 0;
         for(const auto& value : v) {
-            [array addObject:T::Boxed::fromCpp(value)];
+            array[idx++] = T::Boxed::fromCpp(value);
         }
-        return array;
+        NSArray *arrayObject = [NSArray arrayWithObjects:array count:v.size()];
+        free(array);
+        return arrayObject;
     }
```
mknejp: This is not exception safe and leaks memory if an element's fromCpp() throws.

support-lib/objc/DJIMarshal+Private.h
```diff
@@ -229,11 +235,17 @@ class Set {
...
+        NSUInteger idx = 0;
         for(const auto& value : s) {
-            [set addObject:T::Boxed::fromCpp(value)];
+            set[idx++] = T::Boxed::fromCpp(value);
         }
-        return set;
+        NSSet *setObject = [NSSet setWithObjects:set count:s.size()];
+        free(set);
+        return setObject;
     }
```
mknejp: This is also not exception safe.

support-lib/objc/DJIMarshal+Private.h
```diff
@@ -262,11 +274,19 @@ class Map {
...
-            [map setObject:Value::Boxed::fromCpp(kvp.second) forKey:Key::Boxed::fromCpp(kvp.first)];
+            NSUInteger idx = 0;
+            keyObjs[idx] = Key::Boxed::fromCpp(kvp.first);
+            keyObjs[count + idx++] = Value::Boxed::fromCpp(kvp.second);
         }
-        return map;
+        NSDictionary *dictionary = [NSDictionary dictionaryWithObjects:keyObjs+count forKeys:keyObjs count:count];
+        free(keyObjs);
+        return dictionary;
     }
```
mknejp: As isn't this.

mknejp:
> which would bring the public/private ObjC pairs more inline with what people typically use (Where the private header always includes the public one, since it is an extension with added functionality)

These private files are only intended to be used by generated files and generated files only. They aren't ever meant to be used by humans so this doesn't matter. The way it is now aims at minimizing compile times by only including the public headers in the files that actually need them. And just as I'm writing this I realize the line above the one you highlighted should also be a forward declaration instead.

steipete:
I agree - the last commit was half-assed and had several problems - just finished it for the benchmark.

We have a few cases where djinni translation/usage is slow and use the C++ objects directly via the conversion method in the private header. If that's not the intended use, what way would you recommend to access the C++ objects directly? We have cases where it's both nice to have the wrapped ObjC variant around and sometimes necessary to go a level deeper for processing large amounts of objects.

mknejp:
Well, it's not intended, which doesn't mean you can't open the hood and get your hands dirty.

If you are exchanging a lot of data across the language boundary I would first question why it is necessary in the first place to cross the language barrier so frequently. Should you ever intend to port to Android then you don't have this easy access to the C++ world.

If changing the architecture is not an option then you might look into creating your own hand-rolled type that is specialized for the task. For example, if you need to very frequently exchange a million floats between languages you might create a type that is backed by a fixed-sized `NSMutableData` and its content is mapped into C++ as a `array_ref<float>`. Then both sides have read-write access to the same memory region. There is a similar mechanic for Java and JNI. You can then incorporate that type in other Djinni code by using the `@extern` directive.

steipete:
Thanks - appreciate the response.

PR #130, opened 2015-08-14
Objective-C generics

### LINKED ISSUE #103: Generate ObjC generics
Starting in Xcode 7, ObjC supports generics, e.g. `NSArray<NSString *> *foo`, and these translate nicely to Swift. They're not particularly strongly typed but better than nothing. We should optionally emit these.

(Source: https://developer.apple.com/swift/blog/?id=29 and WWDC presentations. Need to find more detailed docs.)

steipete:
This is an attempt to add support for Objective-C generics. Fixes #103.

Note: That's the first time I wrote Scala so this might not be perfect.


mknejp:
This should probably wait until Xcode 7 is out of Beta otherwise people still on 6 won't be able to use the generated code.

PR #131, opened 2015-08-14
Improve Objective-C debug descriptions for record types

steipete:
This is a somewhat crude implementation of `debugDescription` in Objective-C.

We use the trick of boxing all the primitives so emitting the format string is easier.

`debugDescription` is used by lldb when using `po` and makes debugging really really nice.

PR #132, opened 2015-08-14
Adds `toString()` for the Java generator

steipete:
Similar to #131, implements `toString()` for the Java generator for easier debugging.

### DIFF
--- a/src/source/JavaGenerator.scala
+++ b/src/source/JavaGenerator.scala
@@ -340,6 +340,21 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
 
         }
 
+        w.wl
+        w.wl("@Override")
+        w.w("public String toString()").braced {
+          w.w(s"return ").nestedN(2) {
+            w.wl(s""""${self}{" +""")
+            for (i <- 0 to r.fields.length-1) {
+              val name = idJava.field(r.fields(i).ident)
+              val comma = if (i > 0) """"," + """ else ""
+              w.wl(s"""${comma}"${name}=" + ${name} +""")
+            }
+          }
+          w.wl(s""""}";""")
+        }
+        w.wl
+
         if (r.derivingTypes.contains(DerivingType.Ord)) {
           def primitiveCompare(ident: Ident) {
             w.wl(s"if (this.${idJava.field(ident)} < other.${idJava.field(ident)}) {").nested {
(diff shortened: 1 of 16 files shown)


src/source/JavaGenerator.scala
```diff
@@ -340,6 +340,19 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
...
+        w.wl("@Override")
+        w.w("public String toString()").braced {
+          w.w(s"return ").nestedN(2) {
+            w.wl(s"\042${self}{\042 +")
+            for (f <- r.fields) {
+              w.wl(s"\042,\042 + \042${idJava.field(f.ident)}=\042 + ${idJava.field(f.ident)} +")
+            }
+          }
+          w.wl(s"\042}\042;")
+        }
```
mknejp: You can get rid of the unicode stuff by using triple quotes
`s"""foo"x"foo"""` becomes `foo"x"foo`

steipete: I tried that initially, but had troubles with it. Here's what happens when I search/replace for `\042` into `"""`

```
[info] Compiling 4 Scala sources to /Users/steipete/Documents/Projects/djinni/src/target/scala-2.11/classes...
[error] /Users/steipete/Documents/Projects/djinni/src/source/JavaGenerator.scala:347: unclosed string literal
[error]             w.wl(s""""${self}{""" +")
[error]                                    ^
[error] /Users/steipete/Documents/Projects/djinni/src/source/JavaGenerator.scala:349: unclosed string literal
[error]               w.wl(s"""",""" + """${idJava.field(f.ident)}=""" + ${idJava.field(f.ident)} +")
[error]                                                                                            ^
[error] /Users/steipete/Documents/Projects/djinni/src/source/JavaGenerator.scala:352: unclosed string literal
[error]           w.wl(s""""}""";")
[error]                          ^
[error] /Users/steipete/Documents/Projects/djinni/src/source/JavaGenerator.scala:352: ')' expected but ';' found.
[error]           w.wl(s""""}""";")
```

mknejp: That's not how it works. The triple quotes enclose the entire string and allow single quotes inside. Check the example again.

steipete: Got it! Thanks for taking the time to help me on that!

j4cbo:
Awesome. Same requests as for the ObjC `description` change :)

steipete:
done.

PR #133, opened 2015-08-14
No shadow constructor variables

steipete:
adds a _ before variable names to prevent a warning.

Clang emits “Declaration shadows a field of …” when using the same variable name in the constructor. I googled a bit but couldn’t find a best practice besides `new_` which seems excessive long, so I settled with just an underscore.

Before:
screen shot 2015-08-15 at 01 08 39

After:
screen shot 2015-08-15 at 01 10 36

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -175,9 +175,9 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
         // Constructor.
         if(r.fields.nonEmpty) {
           w.wl
-          writeAlignedCall(w, actualSelf + "(", r.fields, ")", f => marshal.fieldType(f.ty) + " " + idCpp.local(f.ident))
+          writeAlignedCall(w, actualSelf + "(", r.fields, ")", f => marshal.fieldType(f.ty) + " " + idCpp.local(f.ident) + "_")
           w.wl
-          val init = (f: Field) => idCpp.field(f.ident) + "(std::move(" + idCpp.local(f.ident) + "))"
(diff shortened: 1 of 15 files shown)


mknejp:
Better use something that does not start with an underscore. The standard reserves all names starting with an underscore followed by an uppercase letter for the implementation, so if a record member variable begins with a capital letter Djinni will generate non-conforming code.

steipete:
How about `n_`? Any preference?

mknejp:
It doesn't really matter as long as it doesn't violate the standard and makes it unambiguous which parameter initializes which member.

steipete:
I settled for `new_` since it's longer but the least confusing.

steipete:
all rebased :)

src/source/CppGenerator.scala
```diff
@@ -66,7 +66,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
       }
     },
     w => {
-      // std::hash specialization has to go *outside* of the wrapNs
+      // std::hash specialization has to go *outside* of the wraps
```
j4cbo: I think this was correct before - `wrapNs` puts a namespace declaration around things, which we don't want here.

steipete: Oh, figured this was a typo. Can undo that if the rest is lgtm

j4cbo:
I think my preferred naming would actually be a *trailing* underscore - very unlikely to collide with user variable names (whereas I could see someone having both `x` and `new_x`), and still safe as far as the standard's reserved names are concerned.

steipete:
addressed feedback.

PR #135, opened 2015-08-21
modify libtextsort.gyp to add ".h .hpp" file to generated project

tobefuturer:
Generated header files are not included in generated project file, but we often need to scan code in header files when we finish project.


j4cbo:
Thanks! Two requests:
- Can you fill in our CLA at https://opensource.dropbox.com/cla/ ?
- The support lib files have changed a bit. I like the use of `find` to locate them, and there are also header files directly under `support-lib` used by both the Java and ObjC libraries.

artwyman:
Alternative changes have already been merged which include headers in generated projects, so this diff is no longer needed.

Issue #136, opened 2015-08-22
Compile java/jni on linux


pwais:
I'm trying to compile (at least for now) just the java/jni unit test and the current dependence on `std::experimental::optional` is a real stickler; the default clang 3.4 package for ubuntu isn't set up to include it.   have anybody ever tried compiling any of djinni on Linux?  are there any suggestions for working around / omitting `std::experimental::optional` ?  or to at least make it ... optional ;) ?  (It's still in review and is not technically required if a djinni IDL doesn't use it).

cc @mknejp

j4cbo:
It's possible to use a standalone implementation of `optional` with Djinni with the `--cpp-optional-header` and `--cpp-optional-template` parameters. There's a good one at https://github.com/akrzemi1/Optional

pwais:
Thanks @j4cbo !  I was able to get all the C++ code compiling fine using that lib.

It seems that OpenJDK 7's `javac` breaks on the non-escaped unicode characters in [ClientInterfaceImpl.java](https://github.com/dropbox/djinni/blob/master/test-suite/handwritten-src/java/com/dropbox/djinni/test/ClientInterfaceImpl.java#L6) unmappable...  would it be worth patching those to escaped unicodes?

I found some other nits with the Makefile that are not portable to Linux (e.g. `-dynamiclib`).  Would there be any interest in a patch that includes a Linux-capable build (for JNI only)?  If so, might using CMake be OK?

Finally, is `-std=c++1y` / C++14 required, or was #80 meant to remove that? (Did the authors decide to drop a requirement of C++14?  If so, is there any interest in making https://github.com/akrzemi1/Optional an official dependency vs having tests use `-std=c++1y` for XCode? AFAICT `std::optional` did not make it in to C++14 anyways ... )

pwais:
derp, patch for the UTF-8 issue: https://gist.github.com/pwais/b595afcd4d896e113ae8

pwais:
FWIW here is a patch with changes I made for Trusty: https://gist.github.com/pwais/e5a1ee3050604dc3f480 (plus using the Optional code that j4cbo recommended). 

It seems the default LLVM package in Utopic includes `<experimental/optional>` and so checking out Optional is not needed.  

One other thing I noticed: Java 8 is now a requirement due to the tests' use of `java.time`.

I'm planning to work on ironing this out into a more portable change.

pwais:
[...]

PR #140, opened 2015-09-08
djinni for linux

pwais:
This PR introduces the following changes:
 * The tests and example run in (Dockerized) Linux (as well as Mac OS X)
 * The (jni) tests now require CMake to support cross-platform builds.  In particular, we leverage CMake's FindJNI to make JNI setup easy.  (NB: the original Makefile isn't portable, only worked on OSX for me).  Hope adding a dependency on CMake isn't a big deal.  It's very easy to install and (in my experience) much better than gyp for non-mobile applications.
 * Docs are edited throughout to include pointers to the new features
 * Includes a few small code changes to the example / test code to fix things that weren't portable.
 * Includes a new `NativeLibLoader` support class to help load jar-packaged native libraries.  `NativeLibLoader` is somewhat similar to One-JAR (and friends) but is a simple one-file utility.  The example and tests use it, but it's an entirely optional dependency.  I think it's worth including because it reduces djinni dependencies and it should be useful in a large number of user use cases.

Hopefully, you can install CMake and run `make example_localhost && cd test-suite && make java` to exercise all the new functionality.

Testing done:
 * Tested using Docker 1.7.1 on OSX 10.9.4
 * There is a script `test-suite/java/docker/run_dockerized_test.sh` that does not run from any `Makefile` by default.  The script runs the djinni tests against a small set of linux platforms.  This test not only ensures portability but also helps document supported platforms.  In the future, it might be nice to test different JVMs using this facility.
 * I initially wanted to include some Microbenchmarks in the example app to e.g. demonstrate/measure the overhead of Java String translation.  Those will need to wait for a later time, but I hope to contribute that code.
 * NB: I only have XCode 6.2 and no Android studio, so I haven't tested those parts. I'm pretty confident in the little changes I made that touch that stuff, but nevertheless I'd appreciate some help testing.

If somebody could please test this on OSX 10.{10,11}, that'd be grand.  I can't upgrade right now.


pwais:
References #136

j4cbo:
This is great, thanks!

I just got back from a week and a half in the desert, so I'm working through a long backlog of stuff (both internal and external) and it may take me a bit to be able to review this. Improved non-Android Linux support is definitely something we want and I hope to pull this in soon.

pwais:
Brilliant!  If you guys have existing use cases, I would be interested to hear about them.

j4cbo:
This looks just about ready to merge. Two requests first:
- Can you rebase against latest master and squash this down to a single commit?
- Please fill out our CLA at https://opensource.dropbox.com/cla/ to make the lawyers happy :)

Thanks!

pwais:
Thanks @j4cbo !
I have:
✓ Signed Dropbox CLA
✓ Squashed and rebased onto dropbox origin/master
✓ Re-run java tests, docker tests, and tested the java `localhost` demo

I noticed that some of the generated iOS example code changed.  I'm on 10.9 and don't have the latest XCode to verify if this is OK.  (I also don't have an Android dev env set up to verify the integrity of the Android tests and example).  Any chance you or somebody else could give the branch a quick whirl?

Makefile
```diff
@@ -15,6 +15,13 @@ clean:
 	git clone https://chromium.googlesource.com/external/gyp.git ./deps/gyp
 	cd deps/gyp && git checkout -q 0bb67471bca068996e15b56738fa4824dfa19de0
 
+# java / jni dependencies
+./deps/java:
+	mkdir -p ./deps/java/test
+	wget http://central.maven.org/maven2/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar -O ./deps/java/test/hamcrest-core-1.3.jar
+	wget http://central.maven.org/maven2/junit/junit/4.11/junit-4.11.jar -O ./deps/java/test/junit-4.11.jar
+	wget http://central.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.0/jsr305-3.0.0.jar -O ./deps/java/jsr305-3.0.0.jar
+
```
j4cbo: So wget doesn't usually exist on OS X. Switching to curl (and https) would be an option, but for the sake of keeping the initial build simple I'd rather just keep these in the repo.

pwais: Oh huh, I had thought XCode installed wget, guess not.  Ok, will fix this.  Technically to redistribute binaries, the individual licenses need to be made available some way.  In the past, I've done something like the notices on http://www.yelp.com/static?p=thirdpartyterms (search for e.g. Eigen / libsvm) for stuff released in a mobile client app and the lawyers were ok with citing the licenses this way.  I'll do something similar for this change and (IMO most importantly) cite where the binaries came from.

pwais:
Updated to re-include jar dependencies in binary form to make build simpler.  Includes credits / notices.

j4cbo:
Good catch, and thanks!

artwyman:
@pwais I just came across the cmake requirement for the first time, and found it quite confusing, though once I figured it out I was able to address it with "brew install cmake".  In the case of the gyp requirement, we have rules in the top-level Makefile to download the necessary gyp version rather than assuming the user has it installed.  Is it possible to do something similar with cmake to avoid this dependency?  I'd prefer to keep the djinni setup as stand-alone as possible.

If it's not possible, I'd like to at least document the dependency in the README, and maybe add a check for it in the makefiles somewhere so we can prompt the user with friendly instructions on what to do.

pwais:
Hi @artwyman sorry you got bitten here!  

I added details to the test README but not the root level one: https://github.com/dropbox/djinni/tree/master/test-suite#testing

cmake typically requires special privs to install... I don't know of a simple, portable way to install it locally (e.g. if it's not already on the system).  Having the build print an error message with installation hints sounds good; I can create a patch to do that.  

Sorry again for the trouble!

artwyman:
@pwais That would be great, thanks.  Is the intent of the "linux" target to be runnable only on Linux?  It failed for me on OSX without docker, and installing that seems not to be enough (it complained about a daemon not running).  Does that need to be added to the pre-requisite setup instructions?

Does the "java" target not run on Linux?  I thought that making it run in both environments was the point of the use of cmake.

Whatever the intended answer is, it would be nice if the definition of the "all" target could be tweaked to automatically include only targets which are appropriate to the local machine.  If that's not all of them, some makefile conditions can be used to alter the list.

pwais:
Hey @artwyman.  Docker on OSX requires either boot2docker or (I believe) Docker Machine, as Docker itself will run inside an emulator, and you interact with it via a client on your mac.  I purposefully left out detailed installation instructions on Docker setup because they'll likely change as Docker evolves, setup is different for Linux, and Docker has had pretty good setup / docs on their site.  The 'documentation' for the Docker requirement is the pointer here: https://github.com/dropbox/djinni/blob/master/test-suite/java/docker/README.md#quickstart

The `java` target did *not* run on Linux prior to my change for a variety of reasons.  But `make java` now runs on Linux (at least inside Docker; I don't have an actual linux box that I can use for testing).  

I think I might have been a bit overzealous in adding `linux` under `all` for the test-suite.  Yes, let's remove that, since it's mainly an optional stress-testing component.  See your point about having `make test` validate for the current platform only.

pwais:
moved to https://github.com/dropbox/djinni/pull/163

PR #147, opened 2015-10-01
updated .gitignore + deleted autogenerated files

skypjack:
Autogenerated files shouldn't be pushed on git. Maybe it's worth removing them.
Updated the .gitignore file accordingly. The same goes without saying for the file local.properties.


j4cbo:
In general we agree. Generated source is not checked in to in the repo for our products, and we recommend against checking it in for other users as well. The Djinni repo *itself* is an exception, though. Including the generated code in the repo and in diffs makes it much easier to see what effect a codegen change has during the code review process.

skypjack:
Besides the patch itself, it looks to me that the .gitignore does not follow the policy.
I mean, the dir `djinni-output-temp` is excluded, while the `generated-src` one is included, anyway both of them contain auto-generated files, so it looks a bit odd, doesn't it?
I guess either the first one should be removed from the .gitignore file or the second one should be added.

j4cbo:
`djinni-output-temp` is an artifact of the build system; `generated-src` is where the files are to be seen by the project.

skypjack:
Ok, I guess I didn't fully get the need for them, but it makes sense.

Anyway, is it possible to update the `.gitignore` file to discard changes after a `make all` (as an example, git is asking me to commit the files `example/generated-src/objc/TXSItemList.h` and `example/generated-src/objc/TXSItemList.mm`) or at least keep updated the `generated-src` directory?
The `make all` generated files that have to be there as far as I've understood, but they are not there and are not excluded by the .gitignore file, moreover the `make clean` command does not take care of them, so it's a bit tricky for the users to understand how to proceed.

I mean, I've understood that's my fault not to have gotten the need for them, by it would avoid future tickets like that one by new users.

j4cbo:
Oops, that's actually my fault then. Looks like a previous commit didn't update the generated code in example/, which I should have caught. If you pull/rebase to the latest master, things should behave properly again :)

(Fixed by babbe4b1ae09ace9b9edbd40b293a1f241116898)

skypjack:
Don't worry, I deserve a beer and nobody will know about that!! :-)
Thank you a lot for the time you spent explaining me the issue.

PR #157, opened 2015-11-05
Potential Android compile fix

pweiskircher:
Without this fix, our Android compile always failed with the following error:
```
In file included from /Users/pat/Work/PSPDFKit/android/pspdfkit/jni/pspdfkit/../../../../core/PSPDFCore/generated-jni/NativeDataDescriptor.cpp:5:
/Users/pat/Work/PSPDFKit/android/pspdfkit/jni/pspdfkit/../../../../core/Vendor/djinni/support-lib/jni/Marshal.hpp:265:15: error: ambiguous conversion for functional-style cast from 'CppType' (aka 'pspdfkit::oxygen::nn<std::__1::shared_ptr<PDFC::DataProvider> >') to 'CppType' (aka 'shared_ptr< ::PDFC::DataProvider>')
   return j ? CppType(T::Boxed::toCpp(jniEnv, j)) : CppType();
```

I'm not quite sure if this is the right way to fix this but it seems to work perfectly!


smarx:

@pweiskircher, it looks like you've already signed the Dropbox CLA. Thanks!

mknejp:
The line as it is now allows for explicit conversions, so removing that might break code for someone else.

To me it looks like a problem with `pspdfkit::oxygen::nn<std::__1::shared_ptr<PDFC::DataProvider> >` as the compiler cannot properly convert it to `shared_ptr< ::PDFC::DataProvider>`. Have you looked at the conversions the compiler is trying and why it finds more than one candidate?

Maybe try replacing `Cpptype(...)` with `CppType{...}` as that turns the functional-style cast to a braced-initializer? That would keep explicit conversions working.

pweiskircher:
I was actually talking with @j4cbo on Slack about this and he said he had to make a similar change on iOS because of some clang problem: https://github.com/dropbox/djinni/blob/master/support-lib/objc/DJIMarshal%2BPrivate.h#L178

mknejp:
Yes, I saw that after looking at the PR. I was worried that the change disables explicit conversions, but `Boxed::toCpp` *should* by design return something that's compatible with `CppType` anyway so I guess it's fine. It may only break with user-provided `nn` types if they lack a implicit conversion to `shared_ptr`.

steipete:
Should we change/add anything? I'd love to see this in master. Thanks!

artwyman:
Sorry, I think this got missed when I took over primary maintainer duties from @j4cbo, so I never realized it was ready and waiting.  I think it seems fine, and matches what's happening in ObjC.  @j4cbo do you have any thoughts?  Otherwise I'll merge this week.

j4cbo:
:+1: if it works with the internal codebase. If there are problems with `nn` compatibility, the first thing I'd try would be to do something like https://github.com/dropbox/djinni/blob/master/support-lib/objc/DJIMarshal%2BPrivate.h#L178 instead of using the ternary operator, but if this works as-is then go for it.

mknejp:
Someone has to do the nitpicking: it's using tabs instead of spaces for indentation.

artwyman:
Good point about the ternary operator possibly leading to more type-ambiguity than the dual returns.
I'll merge this after I have the time to first patch it and try it out with the Dropbox mobile code.

Correction: I'll merge this after the tabs are also converted to spaces (thanks @mknejp).  Or I'll just fix them myself if I end up having to tweak the solution as @j4cbo suggests.

artwyman:
As @j4cbo anticipated, the ternary operator did cause compile failures with this fix in our internal use cases (from a quick scan of the errors it seems to be specifically when mixing use of nn and optional).  The equivalent change with an if and two separate return statements works fine, though.  I'll go ahead and make that fix and push it shortly, so will close out this PR.  Thanks @pweiskircher for diagnosing this issue.

PR #127, opened 2015-08-03
New "flags" archetype for enums oriented towards bitwise flags

mknejp:
This generates enum values representing individual bits instead of
integers. In C++ this also provides overloaded bitwise operators for
convenient use of `enum class` types. ObjC generates `NS_OPTIONS` and in
Java these special enums are supported via `EnumSet`.

The overloaded operators are `&` `|` `^` `&=` `|=` `^=` `~`.

One can optionally mark enum values with the tags `all` or `none` to generate
special values. `none` creates a zero value and `all` a bitwise-or
combination of all other values. These are absent form Java because
`EnumSet` already provides the necessary facility.

```
my_flags = flags {
  flag1;
  flag2;
  flag3;
  no_flags = none;
  all_flags = all;
}
```

``` cpp
enum class my_flags : unsigned {
  no_flags = 0,
  flag1 = 1 << 0,
  flag2 = 1 << 1,
  flag3 = 1 << 2,
  all_flags = 0 | flag1 | flag2 | flag3,
};

constexpr my_flags operator & | ^ (my_flags lhs, my_flags rhs) noexcept { ... }
constexpr my_flags& operator &= |= ^= (my_flags& lhs, my_flags rhs) noexcept { ... }
constexpr my_flags operator ~ (my_flags x) noexcept { ... }
```

``` objc
typedef NS_OPTIONS(NSUInteger, MyFlags) {
  MyFlagsNoFlags = 0,
  MyFlagsFlag1 = 1 << 0,
  MyFlagsFlag2 = 1 << 1,
  MyFlagsFlag3 = 1 << 2,
  MyFlagsAllFlags = 0 | MyFlagsFlag1 | MyFlagsFlag2 | MyFlagsFlag3,
};
```

``` java
public enum MyFlags {
  Flag1,
  Flag2,
  Flag3,
  ;
}
```

All parameters, returns or fields in Java are typed as `EnumSet<MyFlags>`.

PR #159, opened 2015-11-06
improved specialized hash function

skypjack:
The specializations for the function `std::hash` due to the generated `enum`s are difficult to maintain, for updating the underlying type requires several changes to the file `CppGenerator.scala`; moreover, it's not possible to let the users to declare their own underlying types as long as the latter are hardcoded in the generated functions (that would be a good feature indeed, so why not look forward to it?).
The patch introduces `std::underlying_type` to free the generated functions from the actual underlying types of the `enum`s.
To include `type_traits` is mandatory to use `std::underlying_type`.
In addition, the default type for `enum`s has been changed to `unsigned int`, for it looks more suitable to the way `djinni` uses those `enum`s.

The file `.gitignore` has been updated to exclude any file generated by qtcreator.
It would help having those line in that file for future contributions. Thank you.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -54,11 +54,12 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     val self = marshal.typename(ident, e)
 
     if (spec.cppEnumHashWorkaround) {
+      refs.hpp.add("#include <type_traits>") // needed for std::underlying_type
       refs.hpp.add("#include <functional>") // needed for std::hash
     }
 
     writeHppFile(ident, origin, refs.hpp, refs.hppFwds, w => {
-      w.w(s"enum class $self : int").bracedSemi {
+      w.w(s"enum class $self : unsigned int").bracedSemi {
         for (o <- e.options) {
           writeDoc(w, o.doc)
           w.wl(idCpp.enum(o.ident.name) + ",")
@@ -74,8 +75,9 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
           (w: IndentWriter) => {
             w.wl("template <>")
             w.w(s"struct hash<$fqSelf>").bracedSemi {
-              w.w(s"size_t operator()($fqSelf type) const").braced {
-                w.wl("return std::hash<int>()(static_cast<int>(type));")
+              w.wl(s"using underlying_type = typename std::underlying_type<$fqSelf>::type;")
+              w.w(s"std::size_t operator()(const $fqSelf &type) const").braced {
+                w.wl("return std::hash<underlying_type>()(static_cast<underlying_type>(type));")
               }
             }
           }
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,11 @@
 .DS_Store
 
+# QtCreator stuff
+*.config
+*.creator
+*.creator.user
+*.includes
+
 **/.idea/workspace.xml
 **/.idea/tasks.xml
 


smarx:

@skypjack, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

skypjack:
Dropbox Contributor License Agreement (Individual) signed. Thank you.

mknejp:
I think part of the reason you cannot change the underlying type is because the enums must be translateable to Objective-C and Java without loss of information. `NS_ENUM` and `NS_OPTION` uses `NS(U)Integer` for storage, which is either 32 or 64 bit depending on the platform. So technically we should probably use exactly sized underlying types for generated enums.

src/source/CppGenerator.scala
```diff
@@ -54,11 +54,12 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     val self = marshal.typename(ident, e)
 
     if (spec.cppEnumHashWorkaround) {
+      refs.hpp.add("#include <type_traits>") // needed for std::underlying_type
       refs.hpp.add("#include <functional>") // needed for std::hash
     }
 
     writeHppFile(ident, origin, refs.hpp, refs.hppFwds, w => {
-      w.w(s"enum class $self : int").bracedSemi {
+      w.w(s"enum class $self : unsigned int").bracedSemi {
```
mknejp: Changing all enums from signed to unsigned is a very stealthy (possibly breaking) change that has nothing to do with improving the hash function.

src/source/CppGenerator.scala
```diff
@@ -74,8 +75,9 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
           (w: IndentWriter) => {
             w.wl("template <>")
             w.w(s"struct hash<$fqSelf>").bracedSemi {
-              w.w(s"size_t operator()($fqSelf type) const").braced {
-                w.wl("return std::hash<int>()(static_cast<int>(type));")
+              w.wl(s"using underlying_type = typename std::underlying_type<$fqSelf>::type;")
+              w.w(s"std::size_t operator()(const $fqSelf &type) const").braced {
```
mknejp: enums should be taken by value, not `const&`

mknejp:
I approve of using `underlying_type` though, that is definitely more future proof.

src/source/CppGenerator.scala
skypjack: well, it is not a problem to get them either by value or reference, for I understood that the type will be `int` at least for a while and copying `int`s around won't hurt anybody. I'm going to rollback it in a few minutes.

mknejp:
Hmm although I now wonder whether we arn't artificially slowing down compile times with `<type_traits>` included in all enum headers. After all, the code generator knows exactly what the underlying type is.

src/source/CppGenerator.scala
skypjack: I was pretty sure that the response would have been that one. :-)
I'm going to rollback it with the next commit. It makes perfectly sense.

skypjack:
I don't have any evidence of significant slow down and it would free the `hash` function from the actual underlying type, thus opening possibilities for future improvements. Decoupling is usually my preferred approach (unless it leads to performance or architectural issues), anyway it's up to you to decide. Let me know if it's worth to update the pull request with a further commit, I'm at your disposal.

mknejp:
If we will be able to change the underlying type in `.djinni` files in the future, the code generator will still know exactly what type it is. Unfortunately `hash` must be specialized for every enum type explicitly so  we can't take any shortcuts there.

I certainly appreciate the idea and if this was a hand-written `hash` specialization I would use `underlying_type`, too, but with autogenerated code we have the freedom to do tricks that would be considered unmaintainable code if hand-written.

Let's see what the others have to say on this.

However if Djinni has shortcomings for your use case feel free to go all the way and implement the "change underlying type" feature. If you can make it work with all platforms without losing any information during translation I'm sure it'd be welcome. If all you need are unsigned enums you might also take a look at #127 which provides that (and some more).

skypjack:
So, why don't use a generic `hash` function for `enum`s like the one below?
It would reduce even more the compile time for obvious reason, wouldn't it?
I can simply define it somewhere in the support directory and that's all. Does it make sense?

    namespace std {
        template<class E>
        struct hash {
            size_t operator()(const E &e) const {
                return std::hash<typename std::underlying_type<E>::type>()(e);
            }
        private:
            using sfinae = typename std::enable_if<std::is_enum<E>::value, E>::type;
        };
    };

mknejp:
First, that doesn't compile. You are redefining `std::hash` which is not allowed.

Second, what's the point? The generator has all the information required, it does not depend on the functionality of `underlying_type` and producing an explicit specialization is the solution with the least amount of dependencies and fastest compile times.

If your objection is code duplication then the correct location to reduce duplication is the generator Scala code because that is what needs to be maintained and supported, not the generated text.

j4cbo:
A few thoughts:
- First of all, remember that this whole hash specialization thing is a workaround for a C++ bug, [LWG2148](http://wg21.cmeerw.net/lwg/issue2148). So if you're only using clang, or your gcc is new enough, go ahead and pass `--cpp-enum-hash-workaround=false` and the whole problem goes away.
- The template version is pretty similar to what libc++ does (libstdc++ is more clever about it). In some sense that's the best way to do it, but outside the stdlib, specializing `std::hash` in a way that doesn't depend on a user-defined type is UB.

Configurable underlying types for enums would be great, but I don't think this is necessary for that. Overall it seems to me that emitting the type twice makes the generated code more clear, rather than less.

And finally, the .gitignore changes should be a separate PR, so I can merge them :)

skypjack:
You are right, I'll create a new PR for the .gitignore. Thank you.
Feel free to close the ticket if you find that there is no room for such a specialization.
At least, I found that I have to increase my knowledge of Scala to fully understand some choices.

artwyman:
I'm not sure I have full context, but I think I'm +1 in favor of implementing the feature you really need, not a different feature which makes your workaround easier.  I'd be in favor of a flag to change the underlying type for enums if you find it useful, and that would have minimal impact on anyone not using the feature.

PR #160, opened 2015-11-09
Support for printing descriptions of non NSObject external types

pweiskircher:
This allows specifying how the description of a external type should be printed.

We wanted to bridge a type to `CGRect`, but ran into the problem of printing the containing types description. A `CGRect`, or any other non NSObject type, can't be printed using `%@` but has to be converted into a `NSObject` first. With this patch, you can simply specify that in the yaml file:

```
   printDescription: 'NSStringFromCGRect(%s)'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ objc:
     boxed: 'MLBRecord1'
     pointer: true
     hash: '%s.hash'
+    printDescription: '%s'
 objcpp:
     translator: '::mylib::djinni::objc::Record1'
     header: '"mylib/djinni/objc/Record1.hpp"'
--- a/example/example.yaml
+++ b/example/example.yaml
@@ -38,6 +38,10 @@ objc:
   # If the type is a "record" and has "eq" deriving then this string must not be empty.
   # It declares a well-formed expression with a single "%s" format placeholder replaced with the variable for which the hash code is needed
   hash: '%s.hash'
+  # Specifies how a description of this type can be printed.
+  # If the type is NOT a subclass of NSObject, this needs to be modified to return a NSObject.
+  # For example, if the type is a CGRect, use `NSStringFromCGRect(%s)`. If the type is a NSInteger, use `@(%s)`.
+  printDescription: '%s'
 objcpp:
   # The fully qualified name of the class containing the toCpp/fromCpp methods.
   translator: '::mylib::djinni::objc::Record1'
--- a/src/source/ObjcGenerator.scala
+++ b/src/source/ObjcGenerator.scala
@@ -403,12 +403,7 @@ class ObjcGenerator(spec: Spec) extends Generator(spec) {
                 case DEnum => w.w(s"@(self.${idObjc.field(f.ident)})")
                 case _ => w.w(s"self.${idObjc.field(f.ident)}")
               }
-              case e: MExtern =>
-                if (e.objc.pointer) {
-                  w.w(s"self.${idObjc.field(f.ident)}")
-                } else {
-                  w.w(s"@(self.${idObjc.field(f.ident)})")
-                }
+              case e: MExtern => w.w(e.objc.printDescription.format("self." + idObjc.field(f.ident)))
               case _ => w.w(s"self.${idObjc.field(f.ident)}")
             }
           }
--- a/src/source/ObjcMarshal.scala
+++ b/src/source/ObjcMarshal.scala
@@ -88,6 +88,11 @@ class ObjcMarshal(spec: Spec) extends Marshal(spec) {
     case e: Enum => "NSNumber"
   }
 
+  def printDescription(td: TypeDecl) = td.body match {
+    case e: Enum => "@(%s)"
+    case _ => "%s"
+  }
+
   // Return value: (Type_Name, Is_Class_Or_Not)
   def toObjcType(ty: TypeRef): (String, Boolean) = toObjcType(ty.resolved, false)
   def toObjcType(ty: TypeRef, needRef: Boolean): (String, Boolean) = toObjcType(ty.resolved, needRef)
--- a/src/source/YamlGenerator.scala
+++ b/src/source/YamlGenerator.scala
@@ -121,7 +121,8 @@ class YamlGenerator(spec: Spec) extends Generator(spec) {
     "header" -> QuotedString(objcMarshal.include(td.ident)),
     "boxed" -> QuotedString(objcMarshal.boxedTypename(td)),
     "pointer" -> objcMarshal.isPointer(td),
-    "hash" -> QuotedString("%s.hash")
+    "hash" -> QuotedString("%s.hash"),
+    "printDescription" -> QuotedString(objcMarshal.printDescription(td))
   )
 
   private def objcpp(td: TypeDecl) = Map[String, Any](
@@ -195,7 +196,8 @@ object YamlGenerator {
       nested(td, "objc")("header").toString,
       nested(td, "objc")("boxed").toString,
       nested(td, "objc")("pointer").asInstanceOf[Boolean],
-      nested(td, "objc")("hash").toString),
+      nested(td, "objc")("hash").toString,
+      nested(td, "objc")("printDescription").toString),
     MExtern.Objcpp(
       nested(td, "objcpp")("translator").toString,
       nested(td, "objcpp")("header").toString),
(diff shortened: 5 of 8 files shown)


smarx:

@pweiskircher, it looks like you've already signed the Dropbox CLA. Thanks!

artwyman:
I like the concept, and can handle the merge.  I'd like someone with more familiarity with external types to review for any issues, though.  @mknejp?

mknejp:
Looks legit.
But it will break already existing YAML files because they won't have the `printDescription` field. Don't know how important that is. And the README needs updating.

pweiskircher:
Whoops, fixed the README.

pweiskircher:
Regarding the existing YAML files: is there an easy way to have an optional field in the YAML file? My scala isn't really that great. But we could assume '%s' if it is not set and it should work for many types.

mknejp:
Don't know about optional fields, didn't go too depp into the YAML framework. Just going `%s` isn't good enough. Extern enums need boxing and we have the `pointer` field that we can use as a heuristics for records.

Or people just have to re-run their generators after updating... oh well.

pweiskircher:
I'm honestly perfectly fine having people re-run their generators after updating. A little bit of work can be expected after updating. But I'm obviously not the one making that decision and I'll gladly look into having a optional field - not sure how quickly I could get to that, though.

mknejp:
I would agree with generated types, but the YAML files are external sources/libraries that people might not be in contol of. Dunno I guess that's a call for @j4cbo and @artwyman. We'll probably going to need some universal solution as I imagine people are going to add more and more fields in the future.

artwyman:
I think it's okay to expect people to re-run the generator after getting a new Djinni version.  I think it's less okay to require them to edit their YAML files or .djinni files after an update.

I think what's needed here is a way for Djinni to deal with missing fields in the YAML by filling in default values.  E.g. the default for printDescription might be "[%s description]".  That's a per-field default which would be known to the generator.  The YAML parser needs to be able to express optional fields to enable that.  I know nothing about YAML so don't know how hard it would be to accomplish this, but it seems like the right feature goal, and I know roughly how I'd do it in XML, protobufs, or JSON.

mknejp:
The problem with default values is that they have to work reliably. A default value is of no use if the next time you run Djinni it generates code that no longer compiles because some external type doesn't have a `-description` method but the code to call it is still created. In cases like this either the entire `-description` method or the offending field better be ditched.

mknejp:
I also think the field in the YAML file should match the generated method name, so if the method is called `description` so should the YAML file just to make it easier to associate.

artwyman:
@mknejp good point.  That might be a case where a feature which depends on non-default custom behavior needs to be opt-in by default.  For the specific case of description, though, don't all ObjC objects have a default implemenation of "description", much like all Java objects have a default toString()?  Or are you thinking of external types which might not be NSObjects?

mknejp:
Yes, the latter, as is the motivation for this PR where they needed a description for `CGRect`. Enums and interfaces can be defaulted in Cocoa and in Java there is nothing other than `enum` and `class` so no problem there. The problem shouldn't be detecting the absence of the YAML field but how to deal with it. In this case I suggest to not print the property description if its not configured for the type. Leave a comment in the source telling the user why in case they scratch their heads about an incomplete description string.

artwyman:
That's a reasonable approach, though I'd recommend printing something, so that it's clear the field is present/known but unprintable.  Alternately, given that the majority of external types will be NSObjects, it might be reasonable to default to calling description, and let it fail to compile for people with custom types without a description.  That's not entirely backward-compatible, but it limits the build breakages to the users who legitimately have some work they need to do (to provide a custom description) rather than everyone with an external type.

mknejp:
> That's not entirely backward-compatible, but it limits the build breakages to the users who legitimately have some work they need to do (to provide a custom description) rather than everyone with an external type.

It doesn't fix the problem that your code suddenly breaks after updating Djinni because some library you're using doesn't have a `description` field for a `struct` in its YAML. The breakage should be on the library author, not the consumer.

artwyman:
True.  It's was a compromise at best.  The fully compatible answer is to never try to call description on a type we don't know has one, as you say.  I think that could be accomplished by using default code for a missing printDescription field which just returned a fixed string like @"<unprintable>"

artwyman:
If only these were C++ types, you could use metaprogramming and SFINAE to do the right thing.  I suspect that ObjC++ isn't smart enough to do the right thing there, though it might be interesting to try.

mknejp:
If the expression doesn't compile you can SFINAE it away, which is useless for `@class` thanks to all the dynamic nonsense and the default `description` is a `+` method. Records are already implemented in .mm files so we have templates to play with.

steipete:
Reading over this (and also since I wrote the code it replaces), here are some ideas to lessen the impact:

*  Alll internal types can be inferred with the previous code by either self.property or @(self.property).
*  This doesn't yet work for structs. There' actually a clang patch coming in the future that will allow @() for structs and automatically wraps this into `NSValue` objects. For now we can introspect the property type and choose this on our own. This is not something pretty, but to give you an idea, I used a similar approach [in my aspect oriented programming library Aspects, which is also pure ObjC.](https://github.com/steipete/Aspects/blob/master/Aspects.m#L748-L780).
*  That would not give us great output though. Using `NSValue` would be the proper way for a method such as `toDict` (unless the primary goal is JSON, then it's again not that useful). For description printing we want something human-readable, and I don't see a way where we can infer this automagically. Even if we detect common things like `CGRect` or `CGSize` this would mean a custom internal mapping and still wouldn't cover all cases.

If we go with optional fields, it's not obvious how we explain this to users that suddenly their code fails (since they use structs). And if the code doesn't fail, the field is missing from description/toDict[toJSON] which could also lead to confusion. While in contract, djinni didn't make any API compatibility guarantees yet and the effort is minimal to update.

steipete:
Can I help to get things moving on this? We're shipping the above code for some time now and it's working out very well for us. The Dropbox app is using it as well :P

mknejp:
Oh, this still exists. Reading over this again just made me realize something. We actually don't necessarily need a new entry in the YAML file if we create a standardized interface for descriptions that relies on C++ function overloading. After all the Objective-C records and interfaces are implemented in `.mm` files, meaning we can make use of overloading.

For example, we might define a default `std::string describe(const T&)` template in `djinni::` for C++ types and a `NSString* describe(T*)` template for Objective-C types. With function overloading and ADL Djinni can just generate
```cpp
NSString* describe(MLBRecord1* x) {
  using djinni::describe;
  NSString* a = describe(x.a);
  NSString* b = describe(x.b);
  return ...;
}
...
- (NSString*)printDescription {
  using djinni::describe;
  return describe(self);
}
```
Including external types would simply need adding an overload for the type in question. The djinni support library then comes shipped with an implementation of `describe()` for the bulit in types and a default returning an empty string.

I know this is quite a drastic deviation from how the PR works, and not a "pure" Objective-C solution, but it works without breaking existing YAML files and it doesn't require implementing the concept of optional YAML field. We basically let the compiler do all the heavy work for us. Adding an existing "pure" Objective-C type to the overload set requires implementing `describe()` somewhere.

guycnicholas:
Before I was directed to this thread I found a comment about boxing and came across this: http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20150720/133941.html
In there isn't it saying that the ability to box things like CGRect should exist?  I know they don't seem to, nor is the objc_boxable attribute available, so I am sort of confused about the issue.

steipete:
That attribute is still very new and also doesn't really help us for a struct -> string conversion. It also requires support from the person who wrote the struct, which might not always be possible.

steipete:
Anything we can do here to move this forward?

artwyman:
I think the author (@pweiskircher) has the next step here.

steipete:
With boxing, this does not seem necessary anymore. @pweiskircher I updated our PSPDFKit-fork, we can close this ticket.

Note to others: SDK 11 allows to box most often-used structs on iOS. A few are notably missing, here's how you can add them:

```
// Allow auto-boxing of common structs used in Foundation, CoreGraphics and UIKit.
typedef struct __attribute__((objc_boxable)) _NSRange NSRange;
typedef struct __attribute__((objc_boxable)) CGAffineTransform CGAffineTransform;
typedef struct __attribute__((objc_boxable)) UIEdgeInsets UIEdgeInsets;
```
(Learn more at https://pspdfkit.com/blog/2017/even-swiftier-objective-c/)

Issue #152, opened 2015-10-28
WARNING in native method: JNI call made without checking exceptions when required


mknejp:
I just upgraded to latest JDK 8 and am getting this warning for basically every C++ -> JNI call when running the test-suite:
```
     [java] .WARNING in native method: JNI call made without checking exceptions when required to from CallIntMethodV
     [java] 	at com.dropbox.djinni.test.TestHelpers.checkSetRecord(Native Method)
     [java] 	at com.dropbox.djinni.test.SetRecordTest.testJavaSetToCppSet(SetRecordTest.java:25)
     [java] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
     [java] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
     [java] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
     [java] 	at java.lang.reflect.Method.invoke(Method.java:497)
     [java] 	at junit.framework.TestCase.runTest(TestCase.java:176)
     [java] 	at junit.framework.TestCase.runBare(TestCase.java:141)
     [java] 	at junit.framework.TestResult$1.protect(TestResult.java:122)
     [java] 	at junit.framework.TestResult.runProtected(TestResult.java:142)
     [java] 	at junit.framework.TestResult.run(TestResult.java:125)
     [java] 	at junit.framework.TestCase.run(TestCase.java:129)
     [java] 	at junit.framework.TestSuite.runTest(TestSuite.java:255)
     [java] 	at junit.framework.TestSuite.run(TestSuite.java:250)
     [java] 	at junit.framework.TestSuite.runTest(TestSuite.java:255)
     [java] 	at junit.framework.TestSuite.run(TestSuite.java:250)
     [java] 	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:84)
     [java] 	at org.junit.runners.Suite.runChild(Suite.java:127)
     [java] 	at org.junit.runners.Suite.runChild(Suite.java:26)
     [java] 	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
     [java] 	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
     [java] 	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
[...]

PR #161, opened 2015-11-11
Add extra exception checks based on warnings

artwyman:
This addresses most but not all of the warnings in #152.  See my notes in the issue for my best theory on the remaining ones.

This appears stable in Dropbox code, but LocalRef leaks can be hard to detect.  @j4cbo @mknejp I could use another set of eyes on this before I merge it.

### DIFF
--- a/support-lib/jni/Marshal.hpp
+++ b/support-lib/jni/Marshal.hpp
@@ -80,7 +80,11 @@ namespace djinni
 		Bool() : Primitive("java/lang/Boolean", "valueOf", "(Z)Ljava/lang/Boolean;", "booleanValue", "()Z") {}
 		friend JniClass<Bool>;
 		friend Primitive<Bool, bool, jboolean>;
-		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept { return jniEnv->CallBooleanMethod(j, method); }
+		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept {
+            const auto result = jniEnv->CallBooleanMethod(j, method);
+            jniExceptionCheck(jniEnv);
+            return result;
+        }
 	};
 	
 	class I8 : public Primitive<I8, int8_t, jbyte>
@@ -88,7 +92,11 @@ namespace djinni
 		I8() : Primitive("java/lang/Byte", "valueOf", "(B)Ljava/lang/Byte;", "byteValue", "()B") {}
 		friend JniClass<I8>;
 		friend Primitive<I8, int8_t, jbyte>;
-		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept { return jniEnv->CallByteMethod(j, method); }
+        static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept {
+            const auto result = jniEnv->CallByteMethod(j, method);
+            jniExceptionCheck(jniEnv);
+            return result;
+        }
 	};
 	
 	class I16 : public Primitive<I16, int16_t, jshort>
@@ -96,7 +104,11 @@ namespace djinni
 		I16() : Primitive("java/lang/Short", "valueOf", "(S)Ljava/lang/Short;", "shortValue", "()S") {}
 		friend JniClass<I16>;
 		friend Primitive<I16, int16_t, jshort>;
-		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept { return jniEnv->CallShortMethod(j, method); }
+        static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept {
+            const auto result = jniEnv->CallShortMethod(j, method);
+            jniExceptionCheck(jniEnv);
+            return result;
+        }
 	};
 	
 	class I32 : public Primitive<I32, int32_t, jint>
@@ -104,7 +116,11 @@ namespace djinni
 		I32() : Primitive("java/lang/Integer", "valueOf", "(I)Ljava/lang/Integer;", "intValue", "()I") {}
 		friend JniClass<I32>;
 		friend Primitive<I32, int32_t, jint>;
-		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept { return jniEnv->CallIntMethod(j, method); }
+        static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept {
+            const auto result = jniEnv->CallIntMethod(j, method);
+            jniExceptionCheck(jniEnv);
+            return result;
+        }
 	};
 	
 	class I64 : public Primitive<I64, int64_t, jlong>
@@ -112,7 +128,11 @@ namespace djinni
 		I64() : Primitive("java/lang/Long", "valueOf", "(J)Ljava/lang/Long;", "longValue", "()J") {}
 		friend JniClass<I64>;
 		friend Primitive<I64, int64_t, jlong>;
-		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept { return jniEnv->CallLongMethod(j, method); }
+        static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept {
+            const auto result = jniEnv->CallLongMethod(j, method);
+            jniExceptionCheck(jniEnv);
+            return result;
+        }
 	};
 
 	class F32 : public Primitive<F32, float, jfloat>
@@ -120,7 +140,11 @@ namespace djinni
 		F32() : Primitive("java/lang/Float", "valueOf", "(F)Ljava/lang/Float;", "floatValue", "()F") {}
 		friend JniClass<F32>;
 		friend Primitive<F32, float, jfloat>;
-		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept { return jniEnv->CallFloatMethod(j, method); }
+        static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept {
+            const auto result = jniEnv->CallFloatMethod(j, method);
+            jniExceptionCheck(jniEnv);
+            return result;
+        }
 	};
 
 	class F64 : public Primitive<F64, double, jdouble>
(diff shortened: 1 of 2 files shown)


support-lib/jni/Marshal.hpp
```diff
@@ -80,7 +80,11 @@ namespace djinni
 		Bool() : Primitive("java/lang/Boolean", "valueOf", "(Z)Ljava/lang/Boolean;", "booleanValue", "()Z") {}
 		friend JniClass<Bool>;
 		friend Primitive<Bool, bool, jboolean>;
-		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept { return jniEnv->CallBooleanMethod(j, method); }
+		static JniType unbox(JNIEnv* jniEnv, jmethodID method, jobject j) noexcept {
```
mknejp: This (and the others below) can no longer be `noexcept` with `jniExceptionCheck()` present

artwyman:
Weird indentation in the diffs appears to be because some of the support lib code is intended with tabs rather than spaces.  Rather than mess with it in this diff I logged #162 and will run a script to convert after this is merged.

support-lib/jni/Marshal.hpp
```diff
@@ -80,7 +80,11 @@ namespace djinni
...
+            const auto result = jniEnv->CallBooleanMethod(j, method);
```
mknejp: `const` here prevents the compiler from moving the return value. Not really important for primitives, but just saying.

support-lib/jni/Marshal.hpp
artwyman: Good catch.  Will fix.

support-lib/jni/Marshal.hpp
```diff
@@ -80,7 +80,11 @@ namespace djinni
...
+            jniExceptionCheck(jniEnv);
```
mknejp: This almost makes me think there should be templated wrappers in the support lib for calling into JNI methods and always combine call + exception check as that pattern is now ubiquitous.

support-lib/jni/Marshal.hpp
artwyman: Yeah, not a worry for primitives.  I did an explicit move of a LocalRef in djinni_support.cpp, but in this case I just followed my default of making everything const if it can be.  I think that RVO should actually make this method neither copy nor move regardless.

support-lib/jni/Marshal.hpp
artwyman: I agree.  A jniCallBooleanMethod(env, obj, args...) wrapper would make this sort of case much less error prone, and perfect-forwarding could make it zero-cost.  I don't want to deal with creating that and modifying all the code to use it at the moment, but I'd welcome it at some point.

support-lib/jni/djinni_support.cpp
```diff
@@ -239,8 +239,13 @@ jint JniEnum::ordinal(JNIEnv * env, jobject obj) const {
...
 LocalRef<jobject> JniEnum::create(JNIEnv * env, jint value) const {
     LocalRef<jobject> values(env, env->CallStaticObjectMethod(m_clazz.get(), m_staticmethValues));
+    jniExceptionCheck(env);
     DJINNI_ASSERT(values, env);
-    return LocalRef<jobject>(env, env->GetObjectArrayElement(static_cast<jobjectArray>(values.get()), value));
+    LocalRef<jobject> result(env,
+                             env->GetObjectArrayElement(static_cast<jobjectArray>(values.get()),
+                                                        value));
+    jniExceptionCheck(env);
+    return std::move(result);
```
mknejp: Never `std::move` local variables in a `return` as it prevents RVO (and borks auto return type deduction). The standard mandates to prefer moving the return value over copying it.

artwyman: Huh, thanks, I learned something today.  I had been assuming RVO would trump the move, but that it would be safer in the case where RVO didn't work (e.g. multiple return values).  I guess that at least in the case of a move-only type like this, the fact that it compiles when I remove the move means it's doing the right thing. :)

Found a good article to clarify:
https://www.ibm.com/developerworks/community/blogs/5894415f-be62-4bc0-81c5-3956e82276f3/entry/RVO_V_S_std_move?lang=en

mknejp: Yeah, that's also the reason why since C++11 one shouldn't automatically declare locals intended to be returned as `const` as then the compiler is forced to actually perform the copy when returning. It can no longer be moved or elided.

artwyman: Good to know.  I would've assumed RVO would take precedence there too, since the compiler knows the const is meaningless, but I see it's a case where the standard specifies behavior in terms of a return "of the same type" so const or rvalue-ref kills it.  That's tricker to get right than I'd have hoped.  And is a reason not to make variables const, which makes me sad.

mknejp: The `const` is actually not meaningless because the compiler cannot know whether the variable where RVO stores the result is declared as `const`. If it isn't then creating a `const` variable in that location would result in UB if the caller accessed it as non-`const`. You can still make locals `const` just prefer not to do so on those you intend to return from the function. And remember RVO only kicks in if you return the *exact same* variable in all control paths, otherwise you only get a move.

artwyman: Yet the opposite case is fine, where the storage location of the return value is declared const, but used as non-const inside the function?  I'm not sure I understand why the non-constness of the return value would matter.  I guess I've thought of const as simply a limitation on behavior.  What are the cases where the compiler would generate something which is actually unsafe so as to need to be UB?  I don't know of any cases where an object's representation or behavior is different because it was constructed const, vs. being constructed non-const, and then accessed through a const reference.  But I'm not a compiler optimization expert, I suppose.

artwyman:
Independent of our side conversation about constness, I think this code is in its final form.  I'll merge tomorrow if no more comments come in.

support-lib/jni/djinni_support.cpp
mknejp:
```cpp
X foo() { const X x; return x; }
int main() {
  X x = foo();
  x.mutate();
}
```
In this case the variable `x` is declared as `const` in the callee. If it used RVO then accessing that variable as non-`const` after the call were UB since you are removing the `const`-ness.

```cpp
X foo() { X x; return x; }
int main() {
  const X x = foo();
  x.access();
}
```
Here `x` is created as non-`const` via RVO and then merely used as `const` in the caller. This is like passing a variable as `const&` argument, but applied locally in the caller.

artwyman: Yeah, I get that.  That's not what I was wondering about.  What I wonder about is what thing the compiler might do in case 1 which would make it unsafe, which motivates the standard being defined that way.  I don't know of any meaningful optimizations which would make that unsafe.  The only ones I can think of (constant value propagation, or overlapping x with some other piece of data of the right value) would have to be disabled by RVO anywa (in the theoretical case where RVO could occur in such a case), since the function foo() has to put X in a location it can't control.  Or put a different way, if mutating a const value could be unsafe, why couldn't the compiler effectively create a non-const xtmp for RVO, and treat x like a const reference to it.  Seems an equally valid response to the situation, since all guaranteed behavior in the language (like const vs. non-const method overloading) isn't based on whether the variable "was created const" but only whether it is "accessed as const".

I might need a compiler expert to answer that, though.  Or it might just be that the standard specified the "same type" limitation specifically so the compiler didn't have to make potentially implementation-dependent decisions of which optimizations take precedence over which others.

The "same type" limitation makes me also wonder whether this might be a legitimate case for declaring a return type of "const X", though I still dislike that (as I similarly dislike arguments of type "const X") because it exposes implementation details to the caller.

j4cbo: Overlapping with something with the same value wouldn't be allowed in general (same reason why the empty struct has size 1 - distinct objects have to have distinct addresses). My goto example for const-related UB is locating constants entirely in ROM, but that doesn't apply here.

The "treat x like a const reference to a non-const temporary" approach sounds like it could work, but there may be something I'm missing.

artwyman: Yeah, "locate in ROM" works for global/static data.  Not generally for local data, unless you had some kind of a pool of ROM with unique addresses but known values (e.g. a pool of zeroes).  But even that wouldn't apply here since RVO says you have to put it in a location specified by the caller, which can't be ROM since the caller has to specify it without knowing your value yet.

mknejp: Variables *declared* as `const` must never be accessed as non-`const`. The reason is that whenever you get a `const&` function argument that variable might be placed in readonly memory. You simply don't know, so mutating it might destroy the world. In case 1 the callee's version is *declared*  as `const`, and since with RVO the variable in the caller is the *same variable* it would mutate a variable originally declared `const`. That violates the rules.

> Or put a different way, if mutating a const value could be unsafe, why couldn't the compiler effectively create a non-const xtmp for RVO, and treat x like a const reference to it.
That is exactly what happens in case 2. But you cannot swap it: it's illegal to create a `const` xtmp and access it as non-`const&` x as would happen in case 1.

artwyman: I'm not sure I buy that "declared as const" is a meaningful thing from a compiler's point of view in this case, given there are 2 variable declarations here anyway.  RVO is collapsing them into one variable, explicitly breaking standard guarantees such as different variables having unique addresses.  It could be allowed to break other guarantees too, if that were deemed to be safe and valuable.

To be clear lest we be arguing across purposes.  I totally buy that using RVO in case 1 is illegal, and don't dispute that.  I'm asking why.  So I'm not looking for an answer in the form saying that it's illegal based on more different ways that it is illegal.  I'm looking for a bits-and-bytes justification for why it should be illegal, because if it wasn't compilers would do something unsafe, or be unable to do something desireable.  Because the standard rules are set up because someone thinks they're a good idea, and if an alternative would be clearly better for developers with no negative side-effects the standard probably would've allowed it.  I kinda want a condensed replay of the discussion in the standards committee which lead to this set of rules.  I suspect nobody in this project has that, though. :)

mknejp: Ok, I did some more research and I stand corrected. Making it const doens't prevent RVO, however in cases where the compiler cannot apply RVO (it has other prerequisites) it is forced to *copy* the return value instead of *move* it.

But RVO is still just an optimization so the code must be well-formed without it. That means making a local of move-only type const prevents you from returning it because it cannot be moved from, making the code ill-formed in the first place. RVO is not allowed to make that magically compile.

artwyman: Ah, thanks for taking the time to research that.  Good to know, and it does make sense that the code has to be legal without RVO, and can't move without RVO.  The standard needs to allow for things like move/copy ctors with unusual semantics or side-effects, and thus can't do substitutions or omissions except where explicitly allowed to.

That makes the best practice a little easier to follow, though still not easy.  For a well-behaved type which is copyable, making it const won't kill RVO (though it's a bit tricky to be sure where RVO will happen).  And for a move-only type, the compiler failure will remind you to remove the const (though an unwary programmer might add a move at the same time like I did).

Good discussion.  I learned a lot.

PR #163, opened 2015-11-11
Some patches for djinni veterans

pwais:
This change:
 * Prints a nice error message if the user is missing cmake.  I tested it by running the build in docker with `cmake` removed
 * Moves the docker tests out of the `all` target for the test suite.  I checked they still work by running `make linux_docker`.
 * Patches docs in a couple of places to make the Docker dependency easier to find.

In response to discussion at: https://github.com/dropbox/djinni/pull/140#issuecomment-155940221

✓ I have previously signed the Dropbox CLA

### DIFF
--- a/test-suite/Makefile
+++ b/test-suite/Makefile
@@ -1,8 +1,8 @@
-.PHONY: all objc java linux
+.PHONY: all objc java linux_docker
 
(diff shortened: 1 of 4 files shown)


smarx:

@pwais, it looks like you've already signed the Dropbox CLA. Thanks!

Issue #164, opened 2015-11-13
How do I specify that record implements an interface?


vi:
```
somerecord = record +c {
    ololo : string;
    qqq : i64;
}

valueable = interface +c {
    getvalue () : i64;
}
```

C++ implementation ties together `somerecord` and `getvalue`.

How do I make `getvalue` callable from Java like a method?

artwyman:
Depending on what you're actually trying to do, the answer is either "unsupported" or "it just works". :)

"Unsupported" answer: There's no way in Djinni to declare that a record implements an interface or the reverse.  They are unrelated types.

"It just works" answer: In C++ specifically, since you're using a custom extended record, you could do it with multiple inheritance:

    class SomeRecord : public SomerecordBase, public Valuable { ... };

Only C++ code can take advantage of that fact, though, and when your record is passed through Djinni to another language it will still be copied into a new record object in that language, which will be unrelated to the interface.

vi:
I'm trying to do this:

```
somerecord = record +c {
    ololo : string;
    qqq : i64;
}

anotherrecord = record +c {
    id : i32;
}

toxml = interface +c {
    toxml(): string;
}

handlerecord = interface +c {
    handle(r : toxml);
    static createmyhandler(): handlerecord;
}
```

```c++
class Somerecord : public SomerecordBase, public Toxml ...;
class Anotherrecord : public AnotherrecordBase, public Toxml ...;
```

```java
Somerecord sr;
Anotherrecord ar;
Handlerecord h;
h.handle(sr);
```

It almost builds, except of Java don't konw that the records are actually also Toxml.

What is the best way to work around or redesign this?

vi:
Found workaround:

```
toxml = interface +c {
    toxml(): string;
    static castFromSomeRecord   (sr : somerecord   ) : toxml;
    static castFromAnotherRecord(ar : anotherrecord) : toxml;
}
```

But that looks inconvenient.

artwyman:
Yeah, that's probably as good a workaround as you're going to find with the current Djinni functionality.  Djinni doesn't know anything about the multiple-inheritance relationships you're setting up in C++, so neither does Java.  An alternative approach is to move away from records entirely and just make an interface with accessors like get_ololo() and get_qqq().  However without interface inheritance that kind of an approach won't scale well to many related types.

PR #166, opened 2015-11-13
Forward declare instead of #include enums/records from interface's hpp

### LINKED ISSUE #165: Fully fledged #includes instead of forward-declared classes are generated
For the second sample from #164, Djinni generates the following header file `toxml.hpp`:

```c++
// This file generated by Djinni from test.djinni

#pragma once

#include "anotherrecord.hpp"
#include "somerecord.hpp"
#include <memory>
#include <string>

namespace qqq {

class Toxml {
public:
    virtual ~Toxml() {}

    virtual std::string toxml() = 0;

    static std::shared_ptr<Toxml> castFromSomeRecord(const Somerecord & sr);

    static std::shared_ptr<Toxml> castFromAnotherRecord(const Anotherrecord & ar);
};

}  // namespace qqq

```

This causes cirtual header inclusing and doesn't build.

I think Djinji should generate forward class declarations instead, like this:

```
class Somerecord;
class Anotherrecord;
```

With this change it builds and successfully runs.

vi:
Fixes #165.

Not tested except of on one sample project mentioned in #164.

---

Note that I never [seriously] learned Scala, also I'm not too familiar with Djinni itself, so don't expect the pull request be a quality one.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -37,12 +37,12 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     var hppFwds = mutable.TreeSet[String]()
     var cpp = mutable.TreeSet[String]()
 
-    def find(ty: TypeRef) { find(ty.resolved) }
-    def find(tm: MExpr) {
-      tm.args.foreach(find)
-      find(tm.base)
+    def find(ty: TypeRef, forwardDeclareOnly: Boolean) { find(ty.resolved, forwardDeclareOnly) }
+    def find(tm: MExpr, forwardDeclareOnly: Boolean) {
+      tm.args.foreach((x) => find(x, forwardDeclareOnly))
+      find(tm.base, forwardDeclareOnly)
     }
-    def find(m: Meta) = for(r <- marshal.references(m, name)) r match {
+    def find(m: Meta, forwardDeclareOnly : Boolean) = for(r <- marshal.references(m, name, forwardDeclareOnly)) r match {
       case ImportRef(arg) => hpp.add("#include " + arg)
       case DeclRef(decl, Some(spec.cppNamespace)) => hppFwds.add(decl)
       case DeclRef(_, _) =>
@@ -130,8 +130,8 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
 
   override def generateRecord(origin: String, ident: Ident, doc: Doc, params: Seq[TypeParam], r: Record) {
     val refs = new CppRefs(ident.name)
-    r.fields.foreach(f => refs.find(f.ty))
-    r.consts.foreach(c => refs.find(c.ty))
+    r.fields.foreach(f => refs.find(f.ty, false))
+    r.consts.foreach(c => refs.find(c.ty, false))
     refs.hpp.add("#include <utility>") // Add for std::move
 
     val self = marshal.typename(ident, r)
@@ -255,11 +255,11 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
   override def generateInterface(origin: String, ident: Ident, doc: Doc, typeParams: Seq[TypeParam], i: Interface) {
     val refs = new CppRefs(ident.name)
     i.methods.map(m => {
-      m.params.map(p => refs.find(p.ty))
-      m.ret.foreach(refs.find)
+      m.params.map(p => refs.find(p.ty, true))
+      m.ret.foreach((x)=>refs.find(x, true))
     })
     i.consts.map(c => {
-      refs.find(c.ty)
+      refs.find(c.ty, true)
     })
 
     val self = marshal.typename(ident, i)
--- a/src/source/CppMarshal.scala
+++ b/src/source/CppMarshal.scala
@@ -32,7 +32,7 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
   override def toCpp(tm: MExpr, expr: String): String = throw new AssertionError("cpp to cpp conversion")
   override def fromCpp(tm: MExpr, expr: String): String = throw new AssertionError("cpp to cpp conversion")
 
-  def references(m: Meta, exclude: String): Seq[SymbolReference] = m match {
+  def references(m: Meta, exclude: String, forwardDeclareOnly: Boolean): Seq[SymbolReference] = m match {
     case p: MPrimitive => p.idlName match {
       case "i8" | "i16" | "i32" | "i64" => List(ImportRef("<cstdint>"))
       case _ => List()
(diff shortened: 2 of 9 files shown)


smarx:

@vi, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

src/source/CppGenerator.scala
```diff
@@ -255,11 +255,11 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
   override def generateInterface(origin: String, ident: Ident, doc: Doc, typeParams: Seq[TypeParam], i: Interface) {
     val refs = new CppRefs(ident.name)
     i.methods.map(m => {
-      m.params.map(p => refs.find(p.ty))
-      m.ret.foreach(refs.find)
+      m.params.map(p => refs.find(p.ty, true))
+      m.ret.foreach((x)=>refs.find(x,true))
```
artwyman: Super-nitpick, but you need a space after the comma.  It only jumps out at me because it's different on two adjacent lines. :)

src/source/CppGenerator.scala
```diff
@@ -37,12 +37,12 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     var hppFwds = mutable.TreeSet[String]()
     var cpp = mutable.TreeSet[String]()
 
-    def find(ty: TypeRef) { find(ty.resolved) }
-    def find(tm: MExpr) {
-      tm.args.foreach(find)
-      find(tm.base)
+    def find(ty: TypeRef, fromInterface: Boolean) { find(ty.resolved, fromInterface) }
```
artwyman: I'd suggest turning this definition around and having the boolean specify what it does, not what it's used for.  I.e. I'd call this "declOnly", or reverse it and call it "needFullDefinition".

src/source/CppMarshal.scala
```diff
@@ -45,9 +45,13 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
     case MSet => List(ImportRef("<unordered_set>"))
     case MMap => List(ImportRef("<unordered_map>"))
     case d: MDef => d.defType match {
-      case DEnum | DRecord =>
+      case DEnum | DRecord => List()
```
artwyman: This might be my Scala knowledge being insufficient, but what does it mean to have the empty list here, followed by the longer statement below?  Best I can guess is that the empty list here is redundant, and ends up being evaluated then ignored.

artwyman:
This looks good.  I made a few suggestions inline.  Also please re-generate the code for the examples and the test suite so we can see what effect this change has on the generated code.  If you're not on Mac and/or not set up for Android development, you should be able to at least run the java unit test (make -C test-suite java) and the localhost example (make example_localhost) which will run the necessary generators and validate the builds.  If you can't run the full test-suite, I can do that for you once you've validated what you can.

artwyman:
Also please sign the CLA, per the instructions from the bot above.

src/source/CppMarshal.scala
vi: It is an experimentation thing creeped into the commit.

vi:
Note: `make java` in `test-suite` fails much earier with this in compared to before. Probably something is wrong.

vi:
Found two problems:

1. When record or enum is used as return type (not just argument) forward declaration is not enough;
2. Maybe enums got predeclared as classes.

artwyman:
Records are classes.  Enums aren't, so they'll need different handling.

Having a class as a return value shouldn't require a full definition in the interface header file.  It will need a full definition at the call site, as well as the implementation site, which might be what's missing?

Enums can be forward-declared, but need to be declared as enums not classes: "enum class Foo;" or "enum class Foo : int;".  It's possible that the existing Scala code hasn't dealt with that before and is emitting "class Foo;" instead.

vi:
Yes, it is emitting `class Foo;` because of there is explicit hard coded `"class"` in my code. I'll try to improve that...

vi:
Fixed minor things (including enums), but the main issue (discriminating usage in return types and in arguments) is not fixed.

The error raised by the test-suite is appropriate:

```
     [exec] /mnt/src/git/djinni/test-suite/handwritten-src/cpp/test_helpers.cpp: In static member function ‘static testsuite::SetRecord testsuite::TestHelpers::get_set_record()’:
     [exec] /mnt/src/git/djinni/test-suite/handwritten-src/cpp/test_helpers.cpp:9:39: error: return type ‘class testsuite::SetRecord’ is incomplete
     [exec]  SetRecord TestHelpers::get_set_record() {
```

artwyman:
I think that's a case where it's the handwritten code which is wrong.  It was relying on the fact that the header file was going to include the full definition of the record, which it didn't really need to do.  The hand-written code should simply add the include which is no longer included in the generated code.
This does mean that this diff will potentially break people's existing code.  However I think it's doing so for good reasons (to reduce unnecessary header dependencies) so I'm okay with it.

vi:
Now I see: 

> handwritten-src

Fixed the test, now it stops in the same place (in Java code) as before the contribution.

src/source/CppGenerator.scala
```diff
@@ -37,12 +37,12 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     var hppFwds = mutable.TreeSet[String]()
     var cpp = mutable.TreeSet[String]()
 
-    def find(ty: TypeRef) { find(ty.resolved) }
-    def find(tm: MExpr) {
-      tm.args.foreach(find)
-      find(tm.base)
+    def find(ty: TypeRef, dontIncludeRecordsAndEnums: Boolean) { find(ty.resolved, dontIncludeRecordsAndEnums) }
```
j4cbo: The `dontInclude` is kinda confusing - how about `forwardDeclareRecordsAndEnums`? Name it for what it does, not for what it doesn't do.

src/source/CppGenerator.scala
```diff
@@ -255,11 +255,11 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
   override def generateInterface(origin: String, ident: Ident, doc: Doc, typeParams: Seq[TypeParam], i: Interface) {
     val refs = new CppRefs(ident.name)
     i.methods.map(m => {
-      m.params.map(p => refs.find(p.ty))
-      m.ret.foreach(refs.find)
+      m.params.map(p => refs.find(p.ty, true))
+      m.ret.foreach((x)=>refs.find(x, true))
```
j4cbo: Are the parens around `x` necessary?

j4cbo:
Thanks! Some comments inline. Two other requests:
- Please regenerate the autogenerated files (`run_djinni.sh` in `example/` and `test-suite/`)
- Please squash down to one commit

src/source/CppGenerator.scala
vi: I don't know, I juse copied from what I found searching for "Scala lambda expression".

vi:
> Please regenerate the autogenerated files

What are they doing under version control in the first place?

j4cbo:
You're not the first person to ask :) In projects using Djinni, they
shouldn't be. In Djinni itself, including the output makes code review
easier, since the effect of a change on the output winds up included in the
diff.

vi:
@j4cbo, Done.

PR #167, opened 2015-11-16
Support lib: target Java 7, not 8

j4cbo:
Avoid using Java 8 features in NativeLibLoader, and add compile checks to keep us at 1.7 (at least for now).

@pwais, can you confirm that this works for you? I think it's identical -- we were passing 1 for `maxDepth` to `Files.walk`, so replacing it with list-one-directory calls should be equivalent, but please check.

More generally, there's a question of which version of Java we should be targeting. On the C++ and ObjC sides, Djinni is aggressive in using the very latest. On the Java side, given the state of Java 8 on Android, I think it's worth being more conservative for now. The changes needed to go back to 7 here aren't extensive or particularly painful. If they were, or if at any point in the future we stand to gain a lot from requiring Java 8, then we should definitely do so. (But when we do, we should enforce that in the build system et al, so that trying to build with 7 results in clear errors rather than small breakage.)

### DIFF
--- a/support-lib/java/com/dropbox/djinni/NativeLibLoader.java
+++ b/support-lib/java/com/dropbox/djinni/NativeLibLoader.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.file.DirectoryStream;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
@@ -30,7 +31,6 @@ import java.nio.file.StandardCopyOption;
 import java.util.Collections;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Stream;
 
 /**
  * Utilities for loading native libraries containing djinni interfaces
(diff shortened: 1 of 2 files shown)


smarx:

@j4cbo, it looks like you've already signed the Dropbox CLA. Thanks!

artwyman:
LGTM, but I'll wait for @pwais to confirm this matches his intent with this code.

pwais:
Thanks you two!  I wholeheartedly support compatibility with Java 7 here.  I originally took the Java 8 route since I had thought parts of the generated Java code required it, or that it might be requested during code review.

LGTM.  The java unit tests exercise this code (`ant run-jar`), so if they pass you should be good to go.  The Docker stuff is all Java 8 but that shouldn't be a problem.

j4cbo:
```
$ ant run-jar
Buildfile: /Users/jacob/Code/djinni/test-suite/java/build.xml

run-jar:
     [java] Nov 16, 2015 4:55:47 PM com.dropbox.djinni.NativeLibLoader loadLibraryJUnit version 4.11
     [java] 
     [java] INFO: Loaded /var/folders/q7/v6grl2s93fz6366tft3bdyz80000gn/T/libDjinniTestNative.dylib2556396197733850936dylib
     [java] ............................
     [java] Time: 0.027
     [java] 
     [java] OK (28 tests)
     [java] 

BUILD SUCCESSFUL
Total time: 0 seconds
```

Issue #174, opened 2015-12-13
Compile error on optional Java interfaces


jefftrull:
I'm having trouble building an interface that takes as a parameter an optional interface:

```
if1 = interface +c {
    method(i : optional<if2>);
}

if2 = interface +j {
}
```

Compiling this results in a type conflict between the jni call (in NativeIf1.cpp) supplying an argument of type `djinni::Optional<std::experimental::optional, ns::NativeIf2>::CppType` and the cpp header definition (in if1.hpp) which gives the parameter type as `const std::shared_ptr<ns::If2>&`.  Am I missing an implicit conversion somewhere?  Thanks.

artwyman:
Looks like a case we missed when adding nn<> support.  Fix coming soon, along with tests to validate.  @j4cbo I'd love to get your eyes on the fix, since you did the nn<> support.

PR #175, opened 2015-12-15
Fix optional interface arguments when not using nn

artwyman:
- Fix optional interface arguments when not using nn<> (issue #174).
- Run Djinni in test suite as a Makefile target, not a variable.
- Fixes to makefiles for Xcode 7.2.

@j4cbo I'd like your eyes on this if possible since you did the original nn support.


artwyman:
Ping @j4cbo now that you seem to be paying attention.  I'd love to get this accepted before I get back to work, since have some more partial diffs behind it.

j4cbo:
LGTM!

PR #178, opened 2015-12-21
Fix crash when marshaling sets to Java

izacus:
Sets of primitive Java types aren't properly boxed when moved to Java which causes crashes. This PR brings Set marshaling to be equivalent to List marshaling with proper boxing and `HashSet` initialization of proper size.


smarx:

@izacus, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

izacus:
CLA taken care of.

artwyman:
Thanks for fixing!

PR #182, opened 2016-01-03
Add an option to enable/disable the forward declaration of records and enums

cfo82:
Hi

I am again working on Djinni with C++ /CX and C++ /CLI to use Djinni together with managed Code and for the Windows Phone.

Unfortunately it seems like the C++ /CLI on Visual Studio 2012 does not like Forward declarations on structs and enums. Therefore I added a command line Switch to the Djinni Source Code to enable/disable the forward declaration Feature.

Feel free to pull this Change into Djinni main. If I've missed something in my implemenation please let me know and I'll fix it.

Later on I plan to give you a pull-request on my current work with the C++ /CLI and the C++ /CX Generator. 

Cheers
Christian

### DIFF
--- a/src/source/CppMarshal.scala
+++ b/src/source/CppMarshal.scala
@@ -47,7 +47,7 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
     case d: MDef => d.defType match {
       case DRecord =>
         if (d.name != exclude) {
-          if (forwardDeclareOnly) {
+          if (forwardDeclareOnly && spec.cppUseForwardDeclarations) {
             List(DeclRef(s"struct ${typename(d.name, d.body)};", Some(spec.cppNamespace)))
           } else {
             List(ImportRef(include(d.name)))
@@ -57,7 +57,7 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
         }
       case DEnum =>
         if (d.name != exclude) {
-          if (forwardDeclareOnly) {
+          if (forwardDeclareOnly && spec.cppUseForwardDeclarations) {
             List(DeclRef(s"enum class ${typename(d.name, d.body)};", Some(spec.cppNamespace)))
           } else {
             List(ImportRef(include(d.name)))
--- a/src/source/Main.scala
+++ b/src/source/Main.scala
@@ -31,6 +31,7 @@ object Main {
     var cppOptionalTemplate: String = "std::optional"
     var cppOptionalHeader: String = "<optional>"
     var cppEnumHashWorkaround : Boolean = true
+    var cppUseForwardDeclarations : Boolean = true
     var cppNnHeader: Option[String] = None
     var cppNnType: Option[String] = None
     var cppNnCheckExpression: Option[String] = None
@@ -122,6 +123,8 @@ object Main {
         .text("The header to use for optional values (default: \"<optional>\")")
       opt[Boolean]("cpp-enum-hash-workaround").valueName("<true/false>").foreach(x => cppEnumHashWorkaround = x)
         .text("Work around LWG-2148 by generating std::hash specializations for C++ enums (default: true)")
+      opt[Boolean]("cpp-use-forward-declarations").valueName("<true/false>").foreach(x => cppUseForwardDeclarations = x)
+        .text("Use forward declarations for records and enums in C++ (default: true)")
       opt[String]("cpp-nn-header").valueName("<header>").foreach(x => cppNnHeader = Some(x))
         .text("The header to use for non-nullable pointers")
       opt[String]("cpp-nn-type").valueName("<header>").foreach(x => cppNnType = Some(x))
@@ -279,6 +282,7 @@ object Main {
       cppOptionalTemplate,
       cppOptionalHeader,
       cppEnumHashWorkaround,
+	  cppUseForwardDeclarations,
       cppNnHeader,
       cppNnType,
       cppNnCheckExpression,
--- a/src/source/generator.scala
+++ b/src/source/generator.scala
@@ -44,6 +44,7 @@ package object generatorTools {
                    cppOptionalTemplate: String,
                    cppOptionalHeader: String,
                    cppEnumHashWorkaround: Boolean,
+                   cppUseForwardDeclarations: Boolean,
                    cppNnHeader: Option[String],
                    cppNnType: Option[String],
                    cppNnCheckExpression: Option[String],


smarx:

@cfo82, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

cfo82:
I just sent the Dropbox CLA. Cheers

mknejp:
I feel like this is going down a rabbit hole. If we start adding these switches to bypass compiler problems it's only a matter of time until we have hundreds of compatibility options. That's highly impractical to have a reliable test suite for, especially if the switches specifically start targeting certain compilers. We'd have to run these tests with those specific compilers for regression checks. That's a maintenance nightmare.

I really have a bad feeling about this. It's creating a precedent for adding complexity I think shouldn't exist. But that's just me.

cfo82:
Do you have a Suggestion what to do instead of removing the Code by using a Compiler Switch? Because it really leads to compiler errors in context with C++ /CLI with those Forward declarations turned on.

j4cbo:
It's very possible that there are other things in Djinni's generated code and support library that VS2012 won't support. Visual Studio 2015 has much better C++11/14 support - would it be possible for you to update to 2015?

cfo82:
Unfortunately its not possible to upgrade to VC2015 for me. But I can test if VC2015 would fix the issue. That might be important for the discussion, right?

Second place I have to test will be in context with the C++ /CX parts. Because there you are locked to VS2013 if you want to develop for Windows Phone 8. At least to my Knowledge. However I got the issue 

Alternatively the Djinni could try to detect that its generating C++ /CLI Code and disable Forward declarations in that case. But to me thats not much better than using the command line switch. Maybe even worse because it also adds more complexity to the code.

artwyman:
@cfo82 can you give some more details on the type of errors you're getting?  Is this really a language difference with C++/CX and C++/CLI, or some underlying issue.  Do those languages require a full definition to declare a return type where C++ doesn't?

I'd certainly prefer to find a solution which doesn't require even more command-line options.  I'm also a bit concerned about going too deep down the road of claiming to support C++/CX and C++/CLI in Djinni, since I have no idea how deep that rabbit hole is likely to go, and it'll be difficult for most Djinni devs to test and validate that functionality when making changes.

cfo82:
@artwyman I really do understand the concerns about the command line Switch. However unfortunately I did not have a better idea in the time being. I'll post a few of the things I ran into. For better understanding here is my build configuration:

1) Static Library with all the Djinni generated C++ Interface sources
2) Static Library with the actual Business Code and the implementation of the Interface
3) A DLL compiled with C++ /CLI linking against the first two static libraries. The DLL Project contains only the Djinni generated C++ CLI files.

And just to say, it should not be missing the include file because I use precompiled Header files and the PCH file includes all the C++ Interface files.

These are the issues:
1) Not as important is the warning I get: warning C4400: 'const lc_interface::EventType &' : const/volatile qualifiers on this type are not supported

2) 3>C:\Program Files (x86)\Microsoft Visual Studio 11.0\VC\include\type_traits(1791): error C3699: '&&' : cannot use this indirection on type 'lc_interface::DayType'
3>          F:\Projekte\github\LadyCycle\cpp\LCEngine\src\lcengine/interface/CalendarItemData.hpp(54) : see reference to function template instantiation 'lc_interface::DayType &&std::move<lc_interface::DayType&>(_Ty) throw()' being compiled
3>          with
3>          [
3>              _Ty=lc_interface::DayType &
3>          ]

which both are for the Enum type in this case. To get the error for the structs would be a bit more time consuming. I'd do that if necessary. Ideas on how to fix this without removing the forward declarations would be very welcome.

And speaking of precompiled Header files above... I could also create a pull request for the Addition of PCH files to the C++ Interface if you'd like.

And last but not least: I got some prototype Code working with C++ /CLI and /CX. However there are still bits missing (e.g. the sophisticated Proxy implementation you have and exceptions). For now I just implemented what I needed to get my Things running. And I did port the sample app to C++ /CX. C++ /CLI and C++ /CX are very important for me. The first release of the app will be for Windows Phone 8. And I use the /CLI for the automated tests of the C# Code on Windows.

artwyman:
Does the diff in this PR actually fix both of the warnings you listed?  I'm a bit confused, since they don't sound like warnings related to missing definitions, so much as cases where C++/CLI (or the compiler version) isn't supporting/understanding aspects of C++ (like the rvalue reference denoted by &&).  This might just be a case of unclear error messages, though.

Separately, I think adding generation of pch files might be a god feature, though I'm not educated enough on them to know if there are any compiler-specific details there.  I'm also not sure that most devs would want to pre-compile all of the Djinni headers, but that doesn't need to keep it from being available as a chioce.

cfo82:
Yes, this Pull-Request fixes all the issues. I got a working build without the Forward declarations. And I just ran Djinni with the compiler switch forward-declarations set to true. Setting it to false will fix the build. My guess is that the 2012 Compiler barely Supports those new C++ constructs and then it somehow is too much for it to generate C++ /CLI Code. But the earliest I got time to try on VS2015 would be the Weekend. Since this involves porting the entire codebase to VS2015...

My implementation of the PCH is currently to just have the Option to specify the Name of the PCH file. My addition to djinni then will insert the #include into every .cpp file. But it does nothing else apart from that. I created the pch.h and pch.cpp files manually. Of course generating these file automatically could be quite cool, but I think that would also include some more discussions because one might want to add more stuff to the PCH files than just the files generated by Djinni...

artwyman:
Oh, you're talking about a feature to insert the pch include into the generated files, not to generate the pch itself.  The former seems more obviously broadly applicable, while the latter is more specific to a particular use case, though both could be good features.  I'd suggest opening separate issues or PRs for either or both of those for discussion.

If the PCH is an alternative fix for your C++/CLI problem, I'd be happier to merge that, since it feels less specific, and more broadly useful.  I can see lots of other reasons for a pch to be desireable.  Since it sounds like you're still in the process of getting things to work with C++/CLI I'd suggest holding off on merging this PR until you have a better idea of what other issues might come up, since some of my hesitancy is because I feel like this might be just the first of many issues, not the only one.  You can gather up whatever fixes you end up needing in your own branch (in a forked repo) then look at merging them when things stabilize.  Let me know if that sounds reasonable.

cfo82:
Yes I was talking about inserting the #include Statement. Nothing more. I can create a separate PR for that one for the discussion. Thats fine.

But I think this PR and the PCH are not related. They are separate issues. The compiler issues above are really only fixed by removing the forward declarations.

As to C++ /CLI and /CX I was hoping that you would open a separate branch somewhen just like the Python implementation is done. My current prototype is just lacking some of the features. And there are some things which could be done in a nicer way. Question is how to proceed on that one? As I understand the mechanism of the pull requests I could just open another one for C++ CLI and CX for the discussion about them, right? And I think they should be part of the same branch/pull request because the implementations are quite similar.

artwyman:
Sorry if I'm misunderstanding again, since the effect of the command-line flag in this PR is to remove hte fortward-decls and include full definitions instead.  Does including the full definition via PCH not fix the problem?  I.e. the root of the problem appears to be the presence of the forward-decls, not the absence of the full definition?  That's even more confusing than before, since i'd expect a forward-declaration of an already-defined type to have no affect at all.

Regarding a branch for C++/CLI, I think that's a good idea, similar to the approach to Python.  If you're the owner/driver of that work, I think that branch belongs in a repo under github.com/cfo82 rather than Dropbox.  (Just click the Fork button at the top of the page to create that.)  That'll allow you to commit to it directly, accept pull requests from others, and resolve issues as you like without being blocked on Dropbox.  Then when it makes sense to merge upstream we can handle that as a pull-request into this repo.  The Python work might be a misleading precedent since it happens to be a forked project lead by Dropbox, who also owns the master Djinni repo.  There are other cases of sub-projects collaborating in other repos, though, like the work on adding a flags type.

cfo82:
I think you got it right now. When the forward declarations are gone the problem disappears. The type is fully defined by including the appropriate header file in both cases. That was confusing to me as well and it took a bit of time to find the Problem... The other interesting thing is that forward declaring classes works. Ist just an issue with the structs and the enums.

I am a bit new to GitHub and the collaboration through it I must admit...

I already forked Djinni into my separate repo as you suggested. Then created the branch for cx/cli (you can see its name). But I created the PR for the forward declarations from that branch and as I understand that branch is now blocked with that PR? I guess I need a separate branch for the C++/CLI implementation?

And how will people know that they can collaborate on C++ /CLI /CX on my fork instead of your main repo? I guess its quite hard to find it since there are so many forks. And would you as the drop box team collaborate on the discussion about the C++ /CLI /CX or am I on my own? To me the question comes down to how much time I invest on that implementation. I could either just keep it to the minimum work which I need for my own project. But that might just be a subset with features missing. Or I'd be willing to invest a bit more time and effort to get something more polished. But it seems that spending time on that is only worth it if there is also a chance that it gets used by other people/pulled into the main Djinni repo somewhen. Because else it would be kind of wasted time.

artwyman:
Weird.  Okay, I think I understand what's happening, but need to give up on trying to understand why, and just chalk it up to weirdness of MSVC or C++/CLI which is outside my expertise.

Hmm, I don't have a lot of experience with cross-repo collaboration in GitHub either so wasn't aware of the PR blocking behavior.  You can likely work around it by closing this PR (if it won't be merged immediately) or by having an "active dev" branch then creating a separate release/merge branch from it when it's time to send a PR.

We don't have a central list of ongoing projects, or coordination of them, though only because nobody has time to do it.  I'd encourage you to join the Slack community mentioned in the README as a place for discussion and to find other developers who might be interested in helping out.  In the abstract, I and Dropbox could give input on work on things like C++/CLI, but since we don't use that language, and have only one active part-time open-source maintainer (me), I don't think we'd have much to add to the process, other than validating its effects on the overall tool if/when it gets merged.  I think it's best to keep the work distributed and let you drive it to meet your needs, or broader needs as you see fit.

I'm not saying you should fork indefinitely.  Supporting more languages is a general goal for Djinni in the long term.  We just don't have a great answer yet to how to manage that, given the overhead each new language puts on maintenance of general features.  That's part of why Python isn't merged yet too.  And I'd like to put off figuring out the open questions there until any new language is a bit more mature.

mknejp:
Just out of curiosity (not sure you mentioned this already): does it work with forward decls but without precompiled headers and the `stdafx.h` include? Maybe it breaks because of precompiled headers not the decls. Forward declaring structs is certainly nothing new to C++11 (it is for enums but according to https://msdn.microsoft.com/en-us/library/hh567368(v=vs.110).aspx VS 2012 fully supports it).

On a side note: A detailed PCH discussion is certainly for a new issue, but just my short thoughts on it: Modules are just around the corner and they should make precompiled headers pretty much obsolete as the compiler has to include all that information in the module metadata. VS 2015 already has an experimental implementation and the Clang folks are actively working on it (don't know about GCC). The new standard comes out (hopefully) next year and by that time the compiler support should be pretty much immediately available. My instincts say we won't be seeing any PCHs after that, so adding anything related to Djinni seems of temporary value to me. Furthermore all compilers already have an option to "force include" headers before compiling a cpp, so a code change (which would be problematic for portability) is not necessary for a project-wide implicit include that doesn't serve correctness.

cfo82:
@artwyman: Thanks. What about if I open a new pull request about C++ /CX /CLI just for the discussion? That way there would be some connection but no need to pull/merge the changes immediately. And it would be visible for the public. And I'll see to join the slack community in the meantime.

@mknejp: I am quite sure that I also tested this without the precompiled header. It took me some time to figure out what the issue really was and I am quite sure that I tried all those combinations. But to be sure I would need to try again.

I guess the discussion with PCH should be in the other PR then. But one point "against" the temporary nature of PCH in relation to the concept of modules is that you would (to my knowledge) block out development of e.g. Windows Phone 8 apps when you lock Djinni onto only the most recent compiler. Since that development is still done through VS2012 and VS2013. I know that you seem to have a philosophy of going with the newest tech, but please know that this might be an issue for people relying on "older" things.

artwyman:
Structs are really exactly the same as classes in C++ (just with a different default visibility of members), so if there's a difference in the results of the forward-decl, it's probably tied to the fact that Djinni returns structs (records) by value, while it returns classes (interfaces) by pointer.

If you want a place just for discussion, but the changes aren't finalized yet, I'd suggest opening an issue rather than a PR.  And you can link to this PR in the issue so the discussion so far can be accessed.

I wouldn't count on modules to be widespread particularly soon, so I wouldn't hold off on useful features on their account.  I'm not familiar with the compiler options for auto-including headers, but I have found myself considering adding the ability to get Djinni-generated code to include a user-defined  header before, so I could do things like override the behavior of exception-catching macros.  It feels like enough of a generally useful feature that I'd support including it, if someone actually needs it enough to take the time to implement it.

artwyman:
Closing out the question of general C++/CLI support until that's had more discussion, and there's a new PR ready for more complete support.

PR #185, opened 2016-01-10
Include typeid tags as keys in the proxy cache map.

### LINKED ISSUE #179: ObjcProxyCache fails to distinguish two protocols implemented in the same class
## Description
When implementing two `+o` interface types in the same Objective-C class and then storing the same Objective-C object in C++ variables of both types, C++ code that calls methods of the `+o` interfaces will always call the methods of the first interface type that it has seen.

I don't know how I should make this description clearer because I didn't understand the failure mode completely, but please have a look at the example below which should make things clear.

## Affected versions of Djinni
This bug has been introduced with revision b0e3923 and is still present in the current master (163fbe4). Revision 2895f8f is the last version that works fine.

## Example
I have the following simple Djinni IDL file (scroll to the bottom of this post for a link to the full example project):

    FooListener = interface +o {
        foo();
    }

    BarListener = interface +o {
        bar();
    }

    Caller = interface +c {
        static initWithFooListener(fooListener: FooListener, barListener: BarListener): Caller;
        callFoo();
        callBar();
    }

The implementation of `Caller` is straightforward:

    class CallerImpl : public Caller {
    public:
        CallerImpl(const std::shared_ptr<FooListener> &fooListener, const std::shared_ptr<BarListener> &barListener)
        : fooListener(fooListener), barListener(barListener)
        {}

        void callFoo() override {
            fooListener->foo();
        }

        void callBar() override {
            barListener->bar();
        }

    private:
        std::shared_ptr<FooListener> fooListener;
        std::shared_ptr<BarListener> barListener;
    };

    std::shared_ptr<Caller> Caller::initWithFooListener(const std::shared_ptr<FooListener> &fooListener, const std::shared_ptr<BarListener> &barListener) {
        return std::make_shared<CallerImpl>(fooListener, barListener);
    }

I then implemented `FooListener` and `BarListener` in the same Objective-C class:

    @interface Listener : NSObject <EXFooListener, EXBarListener>

When I now construct an instance of `Listener`, pass it to the `Caller` initializer and then call `callBar()`, `foo()` is called instead of `bar()`.

Also check out my [full example project](https://github.com/tiwoc/djinni-cache-bug-demo).

j4cbo:
Fixes #179.

### DIFF
--- a/src/source/JNIGenerator.scala
+++ b/src/source/JNIGenerator.scala
@@ -221,9 +221,8 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
         w.wl(s"friend $baseType;")
         w.wl
         if (i.ext.java) {
-          w.wl(s"class JavaProxy final : ::djinni::JavaProxyCacheEntry, public $cppSelf").bracedSemi {
+          w.wl(s"class JavaProxy final : ::djinni::JavaProxyHandle<JavaProxy>, public $cppSelf").bracedSemi {
             w.wlOutdent(s"public:")
-            // w.wl(s"using JavaProxyCacheEntry::JavaProxyCacheEntry;")
             w.wl(s"JavaProxy(JniType j);")
             w.wl(s"~JavaProxy();")
             w.wl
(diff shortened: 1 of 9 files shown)


smarx:

@j4cbo, it looks like you've already signed the Dropbox CLA. Thanks!

j4cbo:
This has been updated from my previous diff to fix the bad whitespace in proxy_cache_impl.hpp

PR #186, opened 2016-01-12
Tests for typeid tags in the proxy cache map

tiwoc:
This adds tests for PR #185 for the case that multiple +o interfaces are implemented by one class. The tests fail without j4cbo's patch, but succeed with it.


smarx:

@tiwoc, it looks like you've already signed the Dropbox CLA. Thanks!

PR #188, opened 2016-01-15
Feature 184

guycnicholas:
In order to remove the hard coded include path of "../" for extended records  a pair of options were added:
cpp-extended-record-include-prefix & objc-extended-record-include-prefix
Since this removes the default ../ path existing projects like dropbox would need to add something like
-objc-extended-record-include-prefix "../"
to their .sh file


smarx:

@guycnicholas, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

guycnicholas:
I have filled out the CLA

src/source/Main.scala
```diff
@@ -161,6 +163,10 @@ object Main {
         .text("The prefix for #include of the main C++ header files from Objective-C++ files.")
       opt[String]("objcpp-include-objc-prefix").valueName("<prefix>").foreach(x => objcppIncludeObjcPrefixOptional = Some(x))
         .text("The prefix for #import of the Objective-C header files from Objective-C++ files (default: the same as --objcpp-include-prefix)")
+      opt[String]("cpp-extended-record-include-prefix").valueName("<prefix>").foreach(cppExtendedRecordIncludePrefix = _)
+        .text("The prefix for #include of the extended record C++ header files from Objective-C++ files.")
```
artwyman: Do you mean C++ files?
I'd probably just say this as "from generated record headers".  Same below.

artwyman:
This looks good.  I made one comment on the docs inline.

I also want to discuss the default value.  It's no problem for us at Dropbox to deal with adding this argument, but I worry about changing the default behavior for other Djinni users who have made the feature work in its current form, and pull a new version.  They might not be as well informed on recent commits as I am.

I think it would be safer to make the default value of this argument be "../" to preserve the old behavior, but allow it to be overridden with any value from the command-line, including an empty string.

guycnicholas:
I had that same concern. I will change it.

mknejp:
As much as I like keeping it backwards compatible, to new users the default value is going to be very perplexing and I foresee many confusing "wtf?" moments coming our way, because, viewed in isolation, the default makes absolutely no bloody sense.

artwyman:
I do agree with that @mknejp.  It's just not clear to me who to optimize for.
I wish we had a more structured release process.  In that case I'd just put a note about this under "breaking changes" in the release notes, and it would be fine.

If we keep the empty-string default, it'll require changes to the test suite, either to go back to the "../" configuration, or to change files or include paths to make files work without it.

guycnicholas:
This is probably my problem, but if I do set the default to ../ then I can't seem to "empty" it.  For example, adding 
--cpp-extended-record-include-prefix "" \ 
to my .sh file ends up doing nothing and I end up with ../ in my file paths.
I don't understand the Scala code enough, but I think this line is ending up doing nothing because the option is essentially empty.
opt[String]("cpp-extended-record-include-prefix").valueName("<prefix>").foreach(cppExtendedRecordIncludePrefix = _)
Do you have a hint as to how I could fix that?

guycnicholas:
I sort of feel that the "right" way to have this work is to have no initial pathing assumptions.  I realize this means breaking some people, but I wonder what they would do when they pull and the code doesn't build.  Would they then look to the commit comments?  I think if I was not in the Slack channel and was just pulling code, I would read through the commit messages, and as such, if the PR commit message was clear, maybe starting with something like, "*** POSSIBLE BREAKING CHANGE ***" it would get noticed.

artwyman:
That form of command-line appears to work.
I tested it out on the cpp-namespace option.  I changed the "var cppNamespace" at the top to give it a non-empty initial value, then I passed this from the command-line:
    --cpp-namespace ""
That worked to make it use no namespace.  Then I remove the option and from my command-line and saw that the value from the var declaration took effect.

artwyman:
I'm actually convinced on the default being empty-string, though.  Let's go with that, but try to make it as clear as possible.

I'd suggest you collapse your commits into a single one with a clear message talking about the potential breaking change, and submit that as a new PR.  I think you'll also have to change the script which builds the test-suite to deal with the breaking change.

PR #190, opened 2016-01-15
root import error

guycnicholas:
If djinni files were at the root level of the project and one file imported another it would fail with a "null/<filename>" not found.  This fix makes the fileParent a function that tests for nullness and just returns the empty string if necessary.

### DIFF
--- a/src/source/parser.scala
+++ b/src/source/parser.scala
@@ -33,20 +33,24 @@ case class Parser() {
 
 val visitedFiles = mutable.Set[File]()
 val fileStack = mutable.Stack[File]()
-def fileParent:String = if (fileStack.top.getParent() != null) return fileStack.top.getParent() + "/" else return ""
 
 private object IdlParser extends RegexParsers {
   override protected val whiteSpace = """[ \t\n\r]+""".r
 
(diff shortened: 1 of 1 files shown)


smarx:

src/source/parser.scala
```diff
@@ -33,6 +33,7 @@ case class Parser() {
 
 val visitedFiles = mutable.Set[File]()
 val fileStack = mutable.Stack[File]()
+def fileParent:String = if (fileStack.top.getParent() != null) return fileStack.top.getParent() + "/" else return ""
```
artwyman: Can this become an inner declaration inside of importFile?
I worry that its behavior is tailored for a specific use case (given the use of filestack.top), so it doesn't really belong as a general method of this class.  But as a local helper it's great.

artwyman:
Looks good, though i made one suggested improvement inline.

guycnicholas:
I didn't know Scala was a language until like an hour ago.  So my initial attempt failed to get one function inside the other.  Now that you "complained" I worked harder :)

artwyman:
LGTM.  Please confirm that all the tests are happy.  Run "make all" from the top-level djinni dir and it should pass (assuming you're on a Mac, and have the tools to build all platforms).

guycnicholas:
make all succeeded with no errors

PR #191, opened 2016-01-16

guycnicholas:
When extending a record using the +c or +o option the resultant Objective C code includes/imports the subclass assuming it is one level above the generated source.  i.e.

	djinni/src/run \


smarx:

guycnicholas:
Tests have been run and none had to be changed as it appears none of the test code for records does any extending.

mknejp:
In that case it'd be nice to have regression tests added for the feature.

guycnicholas:
I am not sure what that would mean here.

artwyman:
I don't think regression tests are necessary.  We don't have an easy test framework for testing different command-line args.  If we had one in place, I'd want a test for this, but without that I don't think this is risky enough require the effort to create one.  I'll review and merge when I have some time.

mknejp:
Well I though having at least one extended record in C++ and ObjC just to make sure the generated code compiles properly would be nice, since apparently we're not testing record extensions at all right now.

artwyman:
Oh, yes, missed some context on what exactly you meant there.
I thought there were extended records in the test suite.  If there aren't, @guycnicholas can you add one to verify this change?  It certainly seems like adding/changing this argument should have an impact on the test-suite.

guycnicholas:
So this is a brain dead example that compiles.  I think I added all the files per the other examples.

guycnicholas:
After doing this I thought it would be better if I moved my handwritten file outside the normal path and then added in the option to ensure that it compiles.  For some reason, I can't seem to get the option to take effect.  I added:
	--cpp-extended-record-include-prefix "../../handwritten-src/cpp_not_in_path" \
to the run_djinni.sh file at the root of test-suite and I end up with the error
    [exec] #include "extended_record.hpp"
which is unexpected, I should have my extension at the front of that include.  I am assuming I must be putting the option in the wrong file or something.

test-suite/djinni/test.djinni
```diff
@@ -49,3 +49,8 @@ test_helpers = interface +c {
 empty_record = record {
     
 }
+
+# Extended record
```
artwyman: I'd suggest breaking this out into its own file like extended_record.djinni
Also you should add an interface which uses this record, and another record which contains this record, to make sure the include is correct in all possible contexts.

test-suite/handwritten-src/cpp/extended_record.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "extended_record_base.hpp"
+
+namespace testsuite {
+	
+	struct ExtendedRecord : public ExtendedRecordBase
```
artwyman: Is this indented with tabs?
4 spaces is the standard in Djinni code.

test-suite/handwritten-src/cpp/extended_record.hpp
```diff
@@ -0,0 +1,11 @@
...
+	{
+		using ExtendedRecordBase::ExtendedRecordBase;
+	};
+}
```
artwyman: Is there an extra null character here or something?  I'm not sure what GitHub is trying to point out.

src/source/ObjcGenerator.scala
```diff
@@ -178,7 +178,7 @@ class ObjcGenerator(spec: Spec) extends Generator(spec) {
     val self = marshal.typename(objcName, r)
 
     refs.header.add("#import <Foundation/Foundation.h>")
-    refs.body.add("!#import " + q(spec.objcIncludePrefix + (if (r.ext.objc) "../" else "") + marshal.headerName(ident)))
+    refs.body.add("!#import " + q(spec.objcIncludePrefix + (if (r.ext.objc) spec.objcExtendedRecordIncludePrefix else "") + marshal.headerName(ident)))
```
artwyman: Here and elsewhere it looks like there are two different prefixes potentially being added, which would be a problem if both of them were set.  Seems like the code should only add one of them, probably preferring the extended-record prefix if set, and using the other one otherwise.

test-suite/handwritten-src/cpp/extended_record.hpp
mknejp: The file doesn't end in a newline (mouse over the symbol), which is recommended practice and can cause all kinds of problems if not followed.

src/source/ObjcGenerator.scala
guycnicholas: Hmm, well I don't know. I didn't change how the code gets prefixed, just what the prefix is.

test-suite/handwritten-src/cpp/extended_record.hpp
guycnicholas: Fixed this.

guycnicholas:
I am doing the breakout and adding the extra tests now.  In creating the test I realized that the code had a problem with the generated jni header includes that I am fixing as well.

src/source/ObjcGenerator.scala
artwyman: Yes, but if someone sets the new argument using an absolute path (or one relative to some known root in the header search path) it'll cause a mess if they also set the objcIncludePrefix.  The old fixed string you replaced was a relative path which can safely be appended to anything.  It appears now that I've looked more carefully that the "../" prefix is really an addition to the normal prefix which would be used to include the same record if it weren't extended.  That makes sense for Dropbox's code layout where "../" was useful and no other custom prefixes are used.  But in a codebase which needs distinct values for options like objcppIncludePrefix, objcppIncludeCppPrefix, objcppIncludeObjcPrefix, etc. it's unlikely to work.

I don't think my previous suggestion would work either in that case, mind you, without a different extended record prefix which corresponds to each existing prefix option, or some sort of smarts which tries to figure out whether the two prefixes are meant to be appended or not.

Hmm, I don't actually have a great suggestion here which isn't super complicated.  I'm okay with taking the feature as-is and simply documenting the behavior as-is, if you don't want to go further.  It's useful in many cases, just not in all.

guycnicholas: I am sort of confused now.  Originally, I set off to do this because the special ../'s that I saw in the code were ending up in my extended header includes.  I didn't, and still don't, fully understand the full scope of what is going on with all this, but in looking at your comment and reading the code further, it seems that objcIncludePrefix could have been all I needed, except for the issue with the ../ and now that I see a comment in issue 193 about it.  I found the change, though it isn't that new, in commit b38789e6f8b0.

So, I think what I may be asking here, is does the change I am trying to make even need to be there?  Should instead, the hard ../ just be removed or changed?

mknejp: Those are two different things. `--objc-include-prefix` specifies the include path Djinni has to generate in order for the compiler to find the Objective-C files *it is generating*. `--objc-extended-record-include-prefix` specifies what include path Djinni has to generate for the compiler to find *your own handwritten code*. Those paths don't have to be related at all.

artwyman: The diff isn't where the "../" was introduced.  It just included it some of the new/changed lines.  It was already there since the introduction of extended records, which I believe was before the first commit on github.  As Miro points out, the existing include prefix args are for different code, and thus distinct from the new arg you're adding.  But since the "../" was always applied after one of the "find the generated code" prefixes, that's where your new arg is going.  That's workable for some use cases, but not all.

I'm going to put some comments from deeper analysis at the top-level of this PR, rather than on this line.  We can continue the discussion there.

artwyman:
The state of the pre-existing code is more of a mess than I expected.  I took a took a quick look through the code, and wrote added some extended records to the test-suite to examine behavior.  It looks like the full set of situations where the "../" prefix is used for a record "my_rec", which is used as a field in "my_rec2", and an argument in "my_interface":

- CppGenerator.generateRecord: This is for generated my_rec_base.cpp to include hand-written my_rec.hpp.  No other prefix is used.  This seems inconsistent.
- There is no code to apply the prefix when one record includes another, so my_rec2_base.hpp includes my_rec.hpp with no prefix.  This seems like a bug.
- There is no use of ../ in JniGenerator, meaning that NativeMyRec.hpp includes my_rec.hpp with no prefix, which seems like a bug, and doesn't compile in the test suite.  Presumably we have an include path in our internal code which covers this so we've never noticed.
- ObjcGenerator.generateRecord: Genreated DBMyRecBase.mm file including hand-written DBMyRec.h file for the same record.  objcIncludePrefix+"../" is used.
- ObjcMarshal.references: This is for generated DBMyRec2Base.h and DBMyInterface.h to include hand-written DBMyRec.h.  objcIncludePrefix+"../" is used.
- ObjcppMarshal.generateRecord: 2 distinct instances, which are for generated DBMyRecBase+Private.h to include hand-written DBMyRec.h and my_rec.hpp.  They use objcppIncludeObjcPrefix+"../" and "objcppIncludeCppPrefix+"../" respectively.

The scenario which I described fails to compile in the test suite due to the bugs and inconsistencies I point out here.  It's clear that the extended record prefix was never thought through holistically, but applied in the ways which made things work for our use case - sorry about that on behalf of Dropbox developers of the past.

It would be nice to have a more holistic approach to this before creating a new feature with breaking changes, which likely involves defining several new prefix arguments to match/replace the existing ones.  I can understand if you don't feel equipped to do that.  I'm seriously considering just removing the support for the "../" entirely so that all codebases are forced to deal with it by either putting their extended code in the same place as the generated code, or putting it in an include path so it'll be found unqualified.  The latter might be workable for the Dropbox internal use cases, but I'm not sure how hard it'll be to retrofit.  The other quick alternative I can see is to make a more limited command-line argument which simply selectively enables or disables the "../" prefix (without supporting any other prefix).  To avoid breaking changes, we could let the argument default to "on".  However, if we want to encourage other developers not to depend on this half-baked feature, we could make the new argument default to off and be considered "deprecated".  Dropbox, and anyone else who has followed suit can keep using it 

I'm open to feedback.  I can probably spend a little bit of time either implementing or reviewing an alternative here, but don't have the time to dive deep on implementing the ideal feature right now.

mknejp:
I think forcing handwritten code into the same directory as generated code is problematic. In my projects the directory containing the generated files is deleted before every run to make sure no longer existing files get removed and don't linger around.

If I understand your analysis we have five cases where the include prefix should come up:

1. C++ `my_record_base` includes `my_record`. This also covers the case `my_other_record` contains `my_record`
2. Objective-C `DBMyRecordBase` includes `DBMyRecord`. This also covers the case `DBMyOtherRecord` contains `DBMyRecord`
3. Objective-C++ `objcpp::my_record` translation class includes `my_record`
4. Objective-C++ `objcpp::my_record` translation class includes `DBMyRecord`
5. JNI `jni::my_record` translation class includes `my_record`

I'm not exactly in favor of having 5 new command line arguments, but I think we need at least 2 to separate C++ and Objective-C as they may not necessarily be in the same place (one team working on the C++ core, the other on the Objective-C layer).

I think the way to go is to have two prefixes (as in this PR) and apply them in all 5 cases, requiring users to place their handwritten code in a location that can be found from all these scenarios by using appropriate values in their search paths.

guycnicholas:
I would say pulling the hard ../ would be the way to go.  I can easily adjust my include paths on both platforms to find the file wherever I choose to put it, my problem with the current code was simply that in my investigation I was unable to add a path in xCode to get the ../'d file to compile.  I believe that once you put the ../ you have essentially fixed the search location to the current directory.

So, I would say, just remove the ../ which would break some people until they added appropriate include paths. That is simple and allows the code placement to be completely under the individual developers control.  It also is not much of a break for developers, if any at all.

mknejp:
I think that's **too** simple.

You have to remember that the compiler would have to find the files Djinni is generating **and** the files you have written by hand with the **same** include prefix. That either means people dump everything in the same folder, or the include prefix is actually **ambiguous**. then you might accidentally create files with the same name, opening the doors for the compiler to just pick one as it goes down its include search paths, possibly picking the wrong one.

If there's one thing I hate in programming it's ambiguity.

artwyman:
Yeah, removing the ../ entirely would only be a stopgap measure, mostly useful for developers who aren't setting any of the other prefix arguments.  That's the case at Dropbox, where we put our "djinni_gen" output directories into the include paths, so we only need the special case for extended records.  The other prefixes were things we added for open-source generality.

Regarding Miro's list of scenarios: I want to note the important point that when you say a single point covers other cases, that's true only in terms of the conceptual use case (i.e. those two cases would need the same prefix), not the source code (which would need to be updated in multiple places, and is currently inconsistent).

I personally think that simplifying down to 2 prefixes is too complex to be a stopgap feature, but too simple to be a general feature.  In general, assuming a developer might put every type of generated and hand-written file in a different place, you need a different prefix to include A from source B, for every combination of A and B.  That's why for generated code we need 3 options just to include things from ObjC++, depending on whether the included file is ObjC++ or ObjC or C++.

I think that tells me the number of options we need is either 0 or 1 or 5.  0 would be if we just require everything to be in the same directory, or the search path.  1 would be some sort of stop-gap to give a choice between current behavior and the behavior from 0.  5 would be a properly general solution, which basically gives you one new prefix argument for each old prefix argument, and the new prefix would entirely replace the old prefix in the case of extended headers.  There are more than 5 old prefix arguments but only 5 of them are relevant to extended headers.  Below I restate those cases (same as Miro's 5 cases) in terms of the type of file (and its location) for input and output:

- C++ includes C++
- ObjC includes ObjC
- ObjC++ includes ObjC
- ObjC++ includes C++
- JNI includes C++

I can't see how you can reduce that to 2 cases without assuming things about the locations of the file types on the left.

mknejp:
I disagree and here's why:

When generated code includes generated code we need more options because the locations for the generated files is configurable and not under our control. So if code is generated in `src/cpp/` and `src/jni/` then the JNI code requires a different prefix to include C++ headers than to include JNI headers.

However the handwritten files are always in the **same** location, regardless whether you include them from C++, ObjC++ or JNI. They don't move around. So the same search path **will** find your handwritten files from C++, ObjC++ and JNI because it's in the same place for all of them. And it has to be because you need the same file in all three cases.

So all we need is to know "where are your C++ record extensions" and "where are your Objective-C record extensions". The location of those files doesn't depend on where Djinni code is generated at.

artwyman:
I think that's true with certain assumptions, but not be true for all use cases, and my argument here is that we should either be doing the bare minimum, or we should be handling all use cases.

If you're using a relative path (like we currently are with "../") then the prefix you need depends both on the source file and the destination file.  I.e. it does depend on the location of the generated files, unless you happen to arrange things to be at the "same depth" to make relative prefixes work for different sources.  You can avoid the dependency if you use prefix paths which are absolute, or based on some common "root" which is in your header search path.  But I think the reason we support so many different prefixes for generated code is that we're trying to let developers lay out their code however they want.

To give a random contrived example, imagine involve putting objc code in /objc and objcpp code in /objc/objcpp (grouping by sub-languages or something), in which case there's no relative path which will let both kinds of code find a header in /cpp.  Maybe I'm trying to be over-general here, but I feel like the cost of full generality isn't significantly higher than partial generality, and I don't see any reason to bias toward some use cases over others.

guycnicholas:
I am going to respectfully disagree with Miro's argument that it's too simple to just remove the ../.  I believe the ambuiguity mentioned is contrived, and if I do have 2 files with the same name and I created both of them, well that's my bad...and maybe I should be punished a bit for doing that.
I really don't think it is overburdensome to ask a developer to add a couple paths to their includes.
The current code forces me into a particular pathing structure, removing the ../ from the path alleviates that.
Lastly, I will say that things like this really need to be discussed in person over a beer :)

mknejp:
> To give a random contrived example, imagine involve putting objc code in /objc and objcpp code in /objc/objcpp (grouping by sub-languages or something), in which case there's no relative path which will let both kinds of code find a header in /cpp.

Under the assumption that we don't just blindly concatenate paths, i.e. either use no path or `--cpp-extended-record-include-prefix`, then the answer is simply `/cpp`. I'm sorry but when setting up header search paths you gotta put in some common sense. Don't make them ambiguous, avoid file-relative paths, have a common include root. Otherwise you're just asking for maintenance nightmares. It also serves portability since it is implementation-defined how relative paths are resolved. Compilers don't always agree where to start searching from if the path is relative.

In my opinion the reason for so many include paths is the fact we have many output locations to avoid name clashes between files (i.e. `my_record.hpp` for C++ and `my_record.hpp` for JNI) and to separate public/private files to reduce interface bloat (that was my reason for separating objc and objcpp, library clients don't need the `+Private` files). If the files aren't in the same place, of course you need custom include prefixes. But that's simply not the case for the handwritten record extensions. They are always in the same place, there is only one file for all C++, JNI and ObjC++ (otherwise you risk ODR violations) and the files are in some well-known static location in the source tree.

>  I believe the ambuiguity mentioned is contrived, and if I do have 2 files with the same name and I created both of them, well that's my bad...and maybe I should be punished a bit for doing that.

If you're the only person on the team, maybe. But maybe a colleague knows nothing about Djinni and just created a `my_record.hpp` file, and suddenly nothing works. Hours get wasted trying to figure out why the compiler doesn't include the file you're expecting. Stupid compilers and their stupid bugs! So why take the punishment when the cause for the problem can be eliminated in the first place?

I have been on projects where this ambiguity happened. Team A crated a sub-folder with the same name as already existed in team B's source tree. Needless to say, team A included a file from team B, which included `foo/file.hpp`, expecting a file from team B, but instead got team A's file. Hilarity ensued and half a day got wasted hunting down non-existing compiler bugs. Had they instead used long includes like `team_b/bar/foo/file.hpp` this would not have happened and all compilers would have agreed what file to look for. Maybe that experience just made me more aware of this problem and cautious when it comes to setting up header search paths.

So to sum up:

* I do expect people to use common sense when setting up their include search paths and not do things that are largely considered bad practice. Maybe my expectations are too high, but I don't see how anyone would be in favor of ambiguities or prefer to keep a setup that is known to cause maintenance headaches in the long run. Make every file have a *unique* include path in the entire project regardless where it's included from. This improves portability as it is implementation-defined how relative paths are resolved, and Djinni is all about portability.
* Under the above assumption all of Djinni generated C++, JNI and ObjC++ should find the same extended record implementation with the same include prefix regardless where they are generated.

guycnicholas:
Ok, well then, where are we now?  It seems there is a disagreement as to how to proceed, and that both methods obviate my PR.  I would then can there be some short term stop-gap solution here that will work until a longer term, better solution is in place, or?

artwyman:
I think Miro and I should probably agree to disagree.  I think our differences have come down to having different sets of limitations we don't mind imposing on Djinni users.  My most recent example case amounted to a case where I claimed that no relative path would work, but Miro thinks devs should just use an absolute path instead, while I'd rather not force that decision.  But for fairness on the flip side, I think it's okay to require devs not to create multiple files with the same name, while Miro worried about ambiguities in exactly that case.  Since neither of us intends to implement this, the points are mostly moot.  Real use cases are what will prove out the suitability of any feature here.

I suggest we focus on something which can work for now, and iterate.  The main high-level thing that changes for me is that I'd rather not make a breaking change in order to introduce a feature which we don't see as stable/final.  If we go back to this change being purely additive, with the behavior remaining unchanged for users who make no changes, I'm less concerned about the feature being "final".  Personally I'd start with a case which let users decide between the current behavior with "../", or a new behavior which used an empty string in the same places.  It sounds like those two options will work for all the real use cases which are at issue here, and further flexibility can be added by someone who actually plans to use it.

Fixing the cases I discovered where the "../" is actually missing is optional.  I'll likely be forced to fix those eventually as we do refactoring of our code internally.

mknejp:
We're going to break code in some way or another, either now or later.

The `../` solution existing now is obviously insufficient otherwise we wouldn't have this PR.

I suggest we start small and add the 2 minimal options as present in this PR, but not concatenate any paths, meaning replace in `ObjcGenerator.scala`
```scala
refs.body.add("!#import " + q(spec.objcIncludePrefix + (if (r.ext.objc) spec.objcExtendedRecordIncludePrefix else "") + marshal.headerName(ident)))
```
with
```scala
refs.body.add("!#import " + q(if (r.ext.objc) spec.objcExtendedRecordIncludePrefix else spec.objcIncludePrefix) + marshal.headerName(ident)))
```
The implementation in `CppGenerator.scala` already works that way.

As far as "real use cases" go, this is one. @guycnicholas has one that needs solving and the two options in this PR apparently were enough to solve it. We accepted this was going to be a breaking change in some way to get rid of the rather awkward `../` default value. If in future someone finds these two options insufficient we can add in the other ones and let them default to the already existing options where unspecified, thus avoiding another breaking change. We have a solution presented here for a real use case we have **right now** that can be expanded on later.

@guycnicholas does the change I proposed above still solve your issue if you adjust the options appropriately?

guycnicholas:
@mknejp I believe it will work fine.  I will make the change test it and update the PR.  One thing I would ask is can we hold off on adding a bunch of extra test cases if I promise to get back to it in the "near" future?  I am a tight on time right now and can't really afford to spend the time on the extended tests.

guycnicholas:
I created a new PR in which i made the change suggested by mknejp and I squashed it down to a single commit.

artwyman:
Sorry for the delay, folks.  I've been too distracted by other things to keep up on Djinni for a little while.
I agree with closing out this PR and referencing it in the new one.
I haven't looked at the new one yet, but I'm okay with the approach Miro suggested most recently.  So long as the 2 new options could be later extended with new ones in a compatible way, then it addresses my concern about insufficient flexibility.

PR #196, opened 2016-01-29
Extended records forward declare

guycnicholas:
For extended records moved the base struct forward declare into the namespace


smarx:

artwyman:
Is there a case which exercises this in the test suite?  If so, please re-generate the test suite so I can see the changes to the generated code.  If not, can you add one, please?

guycnicholas:
Test added.

src/source/CppGenerator.scala
```diff
@@ -154,6 +153,11 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     def writeCppPrototype(w: IndentWriter) {
       writeDoc(w, doc)
       writeCppTypeParams(w, params)
+      if (r.ext.cpp) {
```
artwyman: You're inserting this between the struct declaration and its template parameters an documentation which are above.  I think you want this to be at the top of the writeCppPrototype function.

Alternately, I think you could just add an element to refs.hppFwds, though you'd have to format it so it works outside of the namespace: "namespace foo { struct bar; };"

test-suite/generated-src/cpp/extended_record_base.hpp
```diff
@@ -0,0 +1,29 @@
...
+// This file generated by Djinni from extended_record.djinni
+
+#pragma once
+
+#include <utility>
+
+namespace testsuite {
+
+/** Extended record */
+struct ExtendedRecord; // Requiring extended class
```
artwyman: Here's the awkward insertion location I mentioned in the comment above.

test-suite/djinni/extended_record.djinni
```diff
@@ -0,0 +1,5 @@
+
+# Extended record
+extended_record = record +c {
+    foo: bool;
+}
```
artwyman: If you want to exercise a case where the forward-decl is actually required, you could add a constant of type extended_record here.

artwyman:
Thanks for fixing.  I made a few comments and corrections inline.

test-suite/djinni/extended_record.djinni
guycnicholas: Funny story, but I just tried this and I can't.  Creating a constant then forces the code to generate a .cpp file inside the generated-src/cpp folder, and that file includes the subclass with an include of the form: #include "../extended_record.hpp" which can't be found if I put the handwritten source with the other handwritten source.  If we wait until my other PR is pulled I could probably get this additional test to work.

guycnicholas:
@artwyman I have updated per comments but left the const test out as it requires the other PR 203.

test-suite/djinni/extended_record.djinni
artwyman: Ah, the irony. :)

artwyman:
This one's ready, so I'll go ahead and merge.  You'll need to merge/rebase #203 after that regardless, and you can make them consistent on where the extended_record definition goes at that point, and also add the constant.

PR #199, opened 2016-01-30
Xcode 7.3 adds a warning if we print an Objective-C class as a pointer...

steipete:
… without an explicit cast. (`-Wformat-pedantic`).

This PR fixes this future issue.


smarx:

@steipete, it looks like you've already signed the Dropbox CLA. Thanks!

artwyman:
Looks good.  @steipete can you re-build the example and test suites and commit the resulting changes in the generated code?  I'd like to see the result of the generator changes included in the same PR before merging.

steipete:
I updated the test files and re-ran the test with Xcode 7.3.1 - everything's looking good.

PR #200, opened 2016-01-30
adapt run script for windows: add ability to run .bat file

vmaks:


artwyman:
Looks good.  I don't have a Windows box to test it on, but I'll take your word for it.
Can you remove the extra echos before the commands run?  Those seem like debug prints which users won't want in normal builds.

Also, I'm not sure why the CLA bot didn't trigger, but I don't see your username in the list of developers who've signed the Dropbox CLA.  You'll need to [sign it here](https://opensource.dropbox.com/cla/) before I can merge.

vmaks:
I signed up the Dropbox CLA and and removed the extra echos before the commands run.

artwyman:
Thanks, @vmaks!

PR #202, opened 2016-02-01

guycnicholas:
When extending a record using the +c or +o option the resultant Objective C code includes/imports the subclass assuming it is one level above the generated source. i.e.

djinni/src/run \


smarx:

PR #203, opened 2016-02-01

guycnicholas:
When extending a record using the +c or +o option the resultant Objective C code includes/imports the subclass assuming it is one level above the generated source. i.e.

djinni/src/run \


smarx:

guycnicholas:
Sorry for all the branch creates and deletes, I was trying to get a PR that was a single commit.

src/source/Main.scala
```diff
@@ -161,6 +163,10 @@ object Main {
         .text("The prefix for #include of the main C++ header files from Objective-C++ files.")
       opt[String]("objcpp-include-objc-prefix").valueName("<prefix>").foreach(x => objcppIncludeObjcPrefixOptional = Some(x))
         .text("The prefix for #import of the Objective-C header files from Objective-C++ files (default: the same as --objcpp-include-prefix)")
+      opt[String]("cpp-extended-record-include-prefix").valueName("<prefix>").foreach(cppExtendedRecordIncludePrefix = _)
+        .text("The prefix for #include of the extended record header files from Objective-C++ files.")
```
artwyman: C++ files not Objective-C++ files.

src/source/ObjcMarshal.scala
```diff
@@ -66,7 +66,7 @@ class ObjcMarshal(spec: Spec) extends Marshal(spec) {
         }
       case DRecord =>
         val r = d.body.asInstanceOf[Record]
-        val prefix = if (r.ext.objc) "../" else ""
+        val prefix = if (r.ext.objc) spec.objcExtendedRecordIncludePrefix else ""
         List(ImportRef(q(spec.objcIncludePrefix + prefix + headerName(d.name))))
```
artwyman: Looks like this line will be combining the new prefix with the old objcIncludePrefix.  I think you want to make it use only one or the other, as you did in ObjcGenerator.scala.

src/source/ObjcppGenerator.scala
```diff
@@ -302,8 +302,8 @@ class ObjcppGenerator(spec: Spec) extends Generator(spec) {
     val noBaseSelf = objcMarshal.typename(ident, r) // Used for constant names
     val cppSelf = cppMarshal.fqTypename(ident, r)
 
-    refs.privHeader.add("!#import " + q(spec.objcppIncludeObjcPrefix + (if(r.ext.objc) "../" else "") + headerName(ident)))
-    refs.privHeader.add("!#include " + q(spec.objcppIncludeCppPrefix + (if(r.ext.cpp) "../" else "") + spec.cppFileIdentStyle(ident) + "." + spec.cppHeaderExt))
+    refs.privHeader.add("!#import " + q(spec.objcppIncludeObjcPrefix + (if(r.ext.objc) spec.objcExtendedRecordIncludePrefix else "") + headerName(ident)))
```
artwyman: Another case where I think you're appending two prefixes rather than using one or the other.  It would be worth testing in case where you set the other prefix args to make sure you can see how they interact.

test-suite/djinni/test.djinni
```diff
@@ -49,3 +49,8 @@ test_helpers = interface +c {
 empty_record = record {
     
 }
+
```
artwyman: Need to merge this with #196 which added the same type, but did so in a separate file.

test-suite/handwritten-src/cpp/extended_record.hpp
```diff
@@ -0,0 +1,11 @@
...
+
+#include "extended_record_base.hpp"
+
+namespace testsuite {
+	
+	struct ExtendedRecord : public ExtendedRecordBase
+	{
+		using ExtendedRecordBase::ExtendedRecordBase;
+	};
+}
```
artwyman: No newline at end of file.  (I don't care, but GitHub complains.)

artwyman:
Thanks for sticking with this, and I'm sorry for the delays on my end.
I made some corrections in inline comments.  I'd also suggest that it might be a good idea to set the new flags to non-empty values in the test suite just to force them to be tested.  Something like "../../handwritten-src/cpp/" will likely work.

guycnicholas:
I think I have this thing pretty good now.

src/source/Main.scala
```diff
@@ -161,6 +163,10 @@ object Main {
         .text("The prefix for #include of the main C++ header files from Objective-C++ files.")
       opt[String]("objcpp-include-objc-prefix").valueName("<prefix>").foreach(x => objcppIncludeObjcPrefixOptional = Some(x))
         .text("The prefix for #import of the Objective-C header files from Objective-C++ files (default: the same as --objcpp-include-prefix)")
+      opt[String]("cpp-extended-record-include-prefix").valueName("<prefix>").foreach(cppExtendedRecordIncludePrefix = _)
+        .text("The prefix for #include of the extended record header files from C++ files.")
```
artwyman: Might be worth clarifying in these messages that the distinctinction is about the header type, not the type which is including this.  Specifically I'd extend this like to "extended record C++ header (.hpp) files" and the same with the other arg with "extended record Objective-C header (.h) files"

artwyman:
Functionality looks great.  Can you make one more tweak to the command-line help text, per my comment inline?  After that I'll merge, and do the internal followups to make Dropbox's internal code work with the new arguments.

guycnicholas:
Updated the help text to what I think you wanted.

artwyman:
I'm satisfied.  I want to verify this works in the Dropbox codebase before merging, and I can't do so at the moment because my Mac is being fixed.  I'll get this merged as soon as I get it back and can pull this down to validate.

PR #206, opened 2016-02-07
CppMarshal: Always emit nn header include for interface

### LINKED ISSUE #197: Missing nn header include when implementing a method that has an argument of its own class
Consider the following Djinni IDL:

	Foo = interface +c {
		useFoo(foo: Foo);
	}

	Bar = interface +c {
		useFoo(foo: Foo);
	}

When built with `--cpp-nn-header "<nn_header.hpp>" --cpp-nn-type "nn_type" --cpp-nn-check-expression "nn_check_expression"`, this generates the following headers (comments have been elided):

Foo.hpp
```cpp
#pragma once

#include <memory>

class Foo {
public:
    virtual ~Foo() {}

    virtual void useFoo(const nn_type<Foo> & foo) = 0;
};
```

Bar.hpp
```cpp
#pragma once

#include <memory>
#include <nn_header.hpp>

class Foo;

class Bar {
public:
    virtual ~Bar() {}

    virtual void useFoo(const nn_type<Foo> & foo) = 0;
};
```

Notice that Foo.hpp lacks the `#include <nn_header.hpp>` that is present in Bar.hpp.

tiwoc:
Previously, the nn header was only included when the argument or return
type of an interface's method was not equal to the interface itself.

Fixes #197

### DIFF
--- a/src/source/CppMarshal.scala
+++ b/src/source/CppMarshal.scala
@@ -66,14 +66,14 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
           List()
         }
       case DInterface =>
-        if (d.name != exclude) {
-          val base = List(ImportRef("<memory>"), DeclRef(s"class ${typename(d.name, d.body)};", Some(spec.cppNamespace)))
-          spec.cppNnHeader match {
-            case Some(nnHdr) => ImportRef(nnHdr) :: base
-            case _ => base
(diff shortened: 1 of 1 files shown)


smarx:

@tiwoc, it looks like you've already signed the Dropbox CLA. Thanks!

PR #210, opened 2016-02-29
Fix ObjC interface generated with non-default cpp identifier type formatting

### LINKED ISSUE #137: Generated C++ and ObjC interface with --ident-cpp-type formatting set to non-default fails to compile
An easy way to reproduce is to take the token.djinni from the test-suite:
token = interface +c +j +o {
  whoami() : string;
}

Then modify test-suite/run_djinni.sh to have pass "--ident-cpp-type foo_bar". You will notice that the proxy class in the DBToken+Private.mm generated will be of ObjProxy type, but the toCpp/fromCpp functions will try to get it from the DbxObjcWrapperCache as a "objc_proxy".

My knowledge of scala is limited but it seems in ObjcppGenerator.scala, that it always generates the proxy class name as ObjcProxy, but when it generates the code the DbxObjcWrapperCache lookup, it uses objcppMarshal.helperClass("objc_proxy") which returns objc_proxy incorrectly.

Should that just be ObjcProxy instead of objc_proxy, since it's already being generated as ObjcProxy elsewhere?

jrogers:
Fixes #137

No functionality change if not using "--ident-cpp-type"

### DIFF
--- a/src/source/ObjcppGenerator.scala
+++ b/src/source/ObjcppGenerator.scala
@@ -190,7 +190,6 @@ class ObjcppGenerator(spec: Spec) extends Generator(spec) {
 
       if (i.ext.objc) {
         w.wl
-        val objcExtSelf = objcppMarshal.helperClass("objc_proxy")
         wrapNamespace(w, spec.objcppNamespace, w => {
           w.wl(s"class $helperClass::ObjcProxy final")
           w.wl(s": public $cppSelf")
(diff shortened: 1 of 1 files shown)


artwyman:
Thanks @jrogers.  Before I merge this I need you to sign our CLA here: https://opensource.dropbox.com/cla/

jrogers:
@artwyman done

PR #214, opened 2016-03-10
Fix shadowing of C++ type names by similar method names.

jpetso:
This is easily triggered by the following kinds of methods:

```
interface node {
  # ...
}

interface tree +c {
  # Invalid method:
  # std::shared_ptr<node> node();
  # The template argument refers to the method name, not the type.
  node(): node;

  # Invalid method:
  # void is_any_of(const std::set<std::shared_ptr<node>>& nodes);
  # The other method name still shadows the 'node' type.
  is_any_of(nodes: set<node>): bool;
}
```

(Apologies for the nonsensical example.)

Both are fixed by using the type's fully qualified name
if the scope contains conflicting names/symbols.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -273,6 +273,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     })
 
     val self = marshal.typename(ident, i)
+    val methodNamesInScope = i.methods.map(m => idCpp.method(m.ident))
 
     writeHppFile(ident, origin, refs.hpp, refs.hppFwds, w => {
       writeDoc(w, doc)
@@ -287,8 +288,8 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
         for (m <- i.methods) {
           w.wl
           writeDoc(w, m.doc)
-          val ret = marshal.returnType(m.ret)
-          val params = m.params.map(p => marshal.paramType(p.ty) + " " + idCpp.local(p.ident))
+          val ret = marshal.returnType(m.ret, methodNamesInScope)
+          val params = m.params.map(p => marshal.paramType(p.ty, methodNamesInScope) + " " + idCpp.local(p.ident))
           if (m.static) {
             w.wl(s"static $ret ${idCpp.method(m.ident)}${params.mkString("(", ", ", ")")};")
           } else {
--- a/src/source/CppMarshal.scala
+++ b/src/source/CppMarshal.scala
@@ -5,27 +5,42 @@ import djinni.generatorTools._
 import djinni.meta._
 
 class CppMarshal(spec: Spec) extends Marshal(spec) {
+  // The scopeSymbols parameter accepted by many of these functions describes a Seq of
+  // symbols/names that are declared in the current scope. The TypeRef or MExpr expression
+  // will be fully qualified if it clashes with any of these symbols, even if full qualification
+  // has not been requested.
 
-  override def typename(tm: MExpr): String = toCppType(tm, None)
+  override def typename(tm: MExpr): String = toCppType(tm, None, Seq())
+  def typename(tm: MExpr, scopeSymbols: Seq[String]): String = toCppType(tm, None, scopeSymbols)
+  def typename(ty: TypeRef, scopeSymbols: Seq[String]): String = typename(ty.resolved, scopeSymbols)
   def typename(name: String, ty: TypeDef): String = ty match {
     case e: Enum => idCpp.enumType(name)
     case i: Interface => idCpp.ty(name)
     case r: Record => idCpp.ty(name)
   }
 
-  override def fqTypename(tm: MExpr): String = toCppType(tm, Some(spec.cppNamespace))
+  override def fqTypename(tm: MExpr): String = toCppType(tm, Some(spec.cppNamespace), Seq())
   def fqTypename(name: String, ty: TypeDef): String = ty match {
     case e: Enum => withNs(Some(spec.cppNamespace), idCpp.enumType(name))
     case i: Interface => withNs(Some(spec.cppNamespace), idCpp.ty(name))
     case r: Record => withNs(Some(spec.cppNamespace), idCpp.ty(name))
   }
 
+  def paramType(tm: MExpr, scopeSymbols: Seq[String]): String = toCppParamType(tm, None, scopeSymbols)
+  def paramType(ty: TypeRef, scopeSymbols: Seq[String]): String = paramType(ty.resolved, scopeSymbols)
   override def paramType(tm: MExpr): String = toCppParamType(tm)
   override def fqParamType(tm: MExpr): String = toCppParamType(tm, Some(spec.cppNamespace))
 
+  def returnType(ret: Option[TypeRef], scopeSymbols: Seq[String]): String = {
+    ret.fold("void")(toCppType(_, None, scopeSymbols))
+  }
   override def returnType(ret: Option[TypeRef]): String = ret.fold("void")(toCppType(_, None))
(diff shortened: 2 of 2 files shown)


smarx:

@jpetso, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

jpetso:
Submitted the CLA.

j4cbo:
Awesome! Thanks in particular for the attention to detail of only qualifying the name if there are conflicting things in scope.

I'll let @artwyman do the final review, but one request from me - to make sure this doesn't regress in the future, can you also add one of your example cases to the test suite, and regenerate the generated code there?

jpetso:
I was looking into test cases briefly, but something about my Java environment and the Java 8 Date API was off and I couldn't manage to fix my environment and build it, even with Ubuntu's update-alternatives clearly pointing to Java 8 for both Java and javac.

I was hoping that by including enough information in the commit message, someone with an understanding of the test framework and an environment that makes it work would find it easy enough to add the test case as a follow-up commit.

If the test case is required for the PR to go through, I can give it another try, or maybe try on a different system.

artwyman:
Djinni shouldn't require Java 8 in general, though it should work with Java 8.  I'd be curious to understand your test issue, since we want to make our test suite as painless as possible.

Regardless, I was able to create a test to exercise this, though it requires twisting the naming convention a bit, because our test suite uses different naming styles for classes and methods.  But this example will do it.  Oddly, I see a problem only for the second method.  The first one compiles just fine on its own.

```
Conflict = interface +c {
}
conflict_user = interface +c {
    Conflict(): Conflict;
    conflict_arg(cs: set<Conflict>): bool;
}
```

I'll give your code a look, and patch it in to try it out.

src/source/CppMarshal.scala
```diff
@@ -6,27 +6,38 @@ import djinni.meta._
 
 class CppMarshal(spec: Spec) extends Marshal(spec) {
```
artwyman: The purpose of the scopeNames variable used pervasively here doesn't seem likely to be clear outside the context of this pull request.  It could use a comment somewhere documenting its purpose.  Also maybe a more explicit name like conflictingSymbolsInScope?

src/source/CppMarshal.scala
```diff
@@ -6,27 +6,38 @@ import djinni.meta._
...
+  def typename(tm: MExpr, scopeNames: Seq[String]): String = toCppType(tm, scopeNames, None)
+  def typename(ty: TypeRef, scopeNames: Seq[String]): String = typename(ty.resolved, scopeNames)
   def typename(name: String, ty: TypeDef): String = ty match {
     case e: Enum => idCpp.enumType(name)
     case i: Interface => idCpp.ty(name)
     case r: Record => idCpp.ty(name)
   }
 
-  override def fqTypename(tm: MExpr): String = toCppType(tm, Some(spec.cppNamespace))
+  override def fqTypename(tm: MExpr): String = toCppType(tm, Seq(), Some(spec.cppNamespace))
```
artwyman: Can you place the new argument last, and allow default arguments in more cases?  I feel like it shouldn't be necessary to pass Seq() in so many places which don't have any information about the presence or absence of conflicting names.

src/source/CppGenerator.scala
```diff
@@ -273,6 +273,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     })
 
     val self = marshal.typename(ident, i)
+    val scopeNames = i.methods.map(m => idCpp.method(m.ident))
```
artwyman: Maybe scopeNames should be methodNamesInScope in this case?

src/source/CppMarshal.scala
```diff
@@ -113,8 +124,17 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
...
-  private def toCppType(ty: TypeRef, namespace: Option[String] = None): String = toCppType(ty.resolved, namespace)
-  private def toCppType(tm: MExpr, namespace: Option[String]): String = {
+  private def toCppType(ty: TypeRef, scopeNames: Seq[String] = Seq(), namespace: Option[String] = None): String =
+    toCppType(ty.resolved, scopeNames, namespace)
+
+  private def toCppType(tm: MExpr, scopeNames: Seq[String], namespace: Option[String]): String = {
+    def withNamespace(name: String): String = {
+      val ns = namespace match {
+        case Some(ns) => Some(ns)
+        case None => if (scopeNames.contains(name)) Some(spec.cppNamespace) else None
```
artwyman: There's an assumption here that there are never unqualified symbols which aren't for generated types.  I think that's a safe assumption currently, at least within the C++ generator (JNI and ObjC use plenty of global names), but it worries me a bit, and is worth at least documenting.

artwyman:
Okay, confirmed that your changes pass the normal test suite, and that my new test case compiles successfully after your changes.

I like the feature, though it's quite specialized and obscure, so I made a bunch of suggestions in the code directed at clarity and documentation.  I want to make it as easy as possible for other developers to understand what this code is for, and avoid breaking it in future.  I realize the Djinni codebase in general is kinda messy, with unclear names and minimal documentation.  I'd like to see that get better with new features, rather than remaining the same. :)

src/source/CppMarshal.scala
jpetso: The reason why I didn't include "conflicting" in here is because the set of names that is passed in is (ideally) the whole set of names, and typically most of them are not in conflict. So I guess the correct term would be `potentiallyConflictingSymbolsInScope`, but that's pretty long. Making the variable name much longer also makes it necessary to expand a good number of function definitions from a single line to three, which I thought makes the code less readable as well.

I don't have a good answer to all of these concerns, but if less compact code is not so much of an issue, I think potentiallyConflictingSymbolsInScope is not the worst choice, it's expressive enough to convey at least some of its purpose immediately. I'll add a comment somewhere around the top of the CppMarshal class.

src/source/CppMarshal.scala
jpetso: Sure. I apparently can't add a default argument to the second `toCppType()` because the Scala compiler complains. I'm adding one for `toCppParamType()`.

src/source/CppMarshal.scala
jpetso: Yes. I wanted to keep the patch down to a reasonable size, so I tried avoiding changes to the parts of the code that deal with more than just C++. It's indeed possible that the same issue will show up elsewhere, and in fact I just found a bug in this commit (not handling `DEnum` and `DRecord` names as well).

That said, I don't think this change makes it harder to deal with this in the future, and the general tendency of Java/Obj-C code to capitalize type names while method names are lowercase means that the most frequent occurrence of these errors should be covered by this change.

Anyway yeah, I'll add a comment.

jpetso:
Pushed a new version that hopefully addresses most of these concerns.

I kept the short name for the parameter in `CppMarshal`, renamed `scopeSymbols` from `scopeNames` since you seem to prefer the term `symbol` here. I didn't rename it to `conflictingSymbolsInScope` because of the correctness concern I expressed above - a symbol will conflict only if the to-be-stringified `TypeRef` or `MExpr` is contained in the list of `scopeSymbols`. I also didn't name it `symbolsInScope` because this makes it sound more like "all the symbols accessible in this scope", whereas the parameter actually means "symbols defined as part of this specific scope" (call it `symbolsOfScope` if you will). I'll let you have the last word on the name.

The new version also fixes enums and records, which were left out by the previous version. Let me know what kind of changes you'd still like to see, if any.

artwyman:
This looks good now.  Sorry for the delay in merging.
I'll add the test case in a follow-up commit.

Issue #113, opened 2015-07-08
Allow default constructor for generated classes


tpecholt:
Add new option for records:
xxx = record {
   a: string;
   b: string;
} deriving (**init**)

which will generate default constructor i.e.
xxx() = default; //C++
xxx() {} //java

The reason is sometimes when records contain lot of members it's impractical to use generated constructor with many arguments. And often many of them have sensible defaults anyway (empty strings, vectors, optionals). But because the generated class already contains 1 generated constructor the default constructor is not created. 

If you don't like default constructor for some reason f.e. because it doesn't initialize scalar values to 0 like in java, generate constructor which will value initialize all members i.e.
xxx() : a(), b() {}

This would be very useful addition

mknejp:
In C++ a `record` is a `struct` for which *brace-initialization* already does the same as the generated constructor, so in theory we could even drop the generated constructor. Then the struct has an implicitly generated default constructor if all members are *default-initialized* and users have the option to use `xxx{}` for *zero-initialization* and `xxx()` for *default-initialization*. However that might be a breaking change because the types can no longer be initialized with parens, meaning `xxx(a, b)` suddenly becomes invalid.

> The reason is sometimes when records contain lot of members it's impractical to use generated constructor with many arguments.

I'd say don't use records with too many fields. in such cases I try to group fields together into other records where it makes logical sense and nest them. This doesn't of course work all the time.

davidchappelle:
Not really sure why structs are even being used. Records should just be classes. Setters and getters should also be predefined for all fields. The defaults for copy and move constructors/assignment-ops should also be generated.

davidchappelle:
For reference #216

artwyman:
I like this idea, though I'm not a fan of the "= default" version of this which doesn't guarantee any initialization at all.  I think it would be best to have well defined default initial values for all types (false, zero, empty), and generate initializers for them in all languages.  Alternately, we could consider @mknejp's suggestion of dropping the ctor in C++ entirely and relying on default braced initialization entirely, though that doesn't address the question for other languages.

I also think it would be cool to go further and allow constants to be used as "initializers" for fields which want non-zero initial values (like what we can do in C++11 now), but that goes well beyond the minimum needed here. :)

pwais:
Yea, the generated ctors (and lack of a default) can definitely get annoying quickly.

[...]

PR #216, opened 2016-03-14
Generate default constructors for c++ records

davidchappelle:
Records can exist as fields within classes. When a record has no
default constructor, it prevents us from using records in classes
that have default constructors. Furthermore, even in cases where
non-default constructors are used, it may not be desireable to
initialize records with specific values at a later point in time.
We could use the brace initializers as a "hack" to get around the
lack of a default constructor, but I would consider using those
types of language constructs to be a poor alternative in this case.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -182,6 +182,9 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
           w.wl(s"friend bool operator>=(const $actualSelf& lhs, const $actualSelf& rhs);")
         }
 
+        w.wl
+        w.wl(actualSelf + "() = default;")
+
         // Constructor.
         if(r.fields.nonEmpty) {
           w.wl


smarx:

@davidchappelle, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

davidchappelle:
Just filled out the CLA.

davidchappelle:
I have another case where I am using a record as a template parameter to an object from a third party library (msgpack-c). The lack of a default constructor is causing that code to break. Here endpoint is a generated djinni record:

```
auto candidate = attributes[attribute::candidate].as<endpoint>();
```

When trying to compile I get the following failure:

```
/home/dchappelle/environment/staging/include/msgpack/object.hpp: In instantiation of ‘typename std::enable_if<(! msgpack::v1::has_as<T>::value), T>::type msgpack::v1::object::as() const [with T = endpoint; typename std::enable_if<(! msgpack::v1::has_as<T>::value), T>::type = endpoint]’:
../../discovery/lan/advertised_endpoint_msgpack.cpp:111:75:   required from here
/home/dchappelle/environment/staging/include/msgpack/object.hpp:535:7: error: no matching function for call to ‘endpoint::endpoint()’
     T v;
       ^
```

Here is the msgpack function:

```
template <typename T>
inline typename std::enable_if<!msgpack::has_as<T>::value, T>::type object::as() const {
    T v;
    convert(v);
    return v;
}
```

The msgpack function looks sensible and a perfectly good example of why a default constructor should be added to the generated c++ djinni output for records.

jpetso:
In general, I am skeptical about default constructors because they can leave a record in an invalid state. I don't think this should be the default behavior in any language.

If you determine that the default constructed state is a valid initial state for your particular type, you might want to use the *_base variant of generated records instead, and add a default constructor in your derived class.

As for msgpack-c, they explicitly support converting non-default-constructible objects: https://github.com/msgpack/msgpack-c/wiki/v1_1_cpp_adaptor (look for the as<> function).

davidchappelle:
Thanks @jpetso I'll give that a try.

artwyman:
Thanks for your suggestion!  A few thoughts on the discussion:

I think most well-behaved template libraries, should provide an option to work for objects with no default constructor, so it's good to hear msgpack-c does so.  That being said, a default constructor is a fine thing to want in generated records for convenience.  Extending the record via +c does sound like it would be a workaround here, but would require a lot of hand-written code, which is a shame when we have a perfectly good code-generator.

I'm not a fan of adding the default constructor by default, though, since some people may not want their records in default state.  I'm particularly not a fan of the "= default" version of this which doesn't guarantee any initialization at all.  I'd suggest adding this as an optional feature using "deriving" as suggested in issue #113.  I also think it would be best to have well defined default initial values for all types (false, zero, empty), and generate initializers for them in all languages, not just C++.

I also think it would be cool to go further and allow constants to be used as "initializers" for fields which want non-zero initial values (like what we can do in C++11 now), but that goes well beyond the minimum needed here. :)

artwyman:
Closing this PR for now, with issue #113 tracking the desired feature.  Re-submit if you want to take a stab at a more fleshed out version.

PR #224, opened 2016-04-14
Changed Namespace for CppProxy to global ::djinni in JNIGenerator

Jakouf:
The Compiler had some issues finding the CppProxy, cause we were using foobar::djinni as namespace for generated djinni files. I changed in the JNIGenerator.scala file the namespace for the CppProxy to the global Namespace. Then the compiler could find the CppProxy and everything was alright for us.

The Problem was during compiling for Android with NDK r11b and the shipped clang compiler 3.6.

### DIFF
--- a/example/generated-src/jni/NativeSortItems.cpp
+++ b/example/generated-src/jni/NativeSortItems.cpp
@@ -18,7 +18,7 @@ CJNIEXPORT void JNICALL Java_com_dropbox_textsort_SortItems_00024CppProxy_native
 {
     try {
         DJINNI_FUNCTION_PROLOGUE1(jniEnv, nativeRef);
-        delete reinterpret_cast<djinni::CppProxyHandle<::textsort::SortItems>*>(nativeRef);
+        delete reinterpret_cast<::djinni::CppProxyHandle<::textsort::SortItems>*>(nativeRef);
     } JNI_TRANSLATE_EXCEPTIONS_RETURN(jniEnv, )
 }
 
--- a/src/source/JNIGenerator.scala
+++ b/src/source/JNIGenerator.scala
@@ -334,7 +334,7 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
           }
         }
         nativeHook("nativeDestroy", false, Seq.empty, None, {
-          w.wl(s"delete reinterpret_cast<djinni::CppProxyHandle<$cppSelf>*>(nativeRef);")
+          w.wl(s"delete reinterpret_cast<::djinni::CppProxyHandle<$cppSelf>*>(nativeRef);")
         })
         for (m <- i.methods) {
           val nativeAddon = if (m.static) "" else "native_"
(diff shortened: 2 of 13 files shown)


smarx:

@Jakouf, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

Jakouf:
Signed the Dropbox CLA

smarx:
				
	@Jakouf, thanks for signing the CLA!

artwyman:
Good catch.
@Jakouf Can you please re-generate the code for the example and test suite (just "make all" will cover it") so those are up-to-date with this change?  Generated code is in git in the Djinni repo to aid in code-reviews (even though that wouldn't be a best practice for users of Djinni).
Otherwise this looks good, so I'll merge after that update.

Jakouf:
Added the new generated cpp and header files.

But a other thing just came up. with NDK r12 Beta there are a lot of files missing. I think we should change the stl to c++_static or something like that and also we should change the compiler for android to clang in the example folder. I can open a new merge request.

Jakouf:
pressed the wrong button -.-

artwyman:
I haven't tried out r12 beta yet.  What are the missing files?  Is it because they got rid of GCC?  We're still partway through migrating to r11 internally, and still using gcc 4.9 (now deprecated).  r11 is the first NDK release which I expect might make it feasible to build Dropbox's code with clang.  However I'd be hesitant to make it a default requirement for Djinni until I'm more confident that it's stable.  Feel free to open a separate issue or PR to discuss that.  Maybe we should consider making more NDK settings configurable via env vars, so that developers can make their own choices when building the examples and test suite.

Jakouf:
Looks like he cannot find the stl or the stl is not complete.

`:app:ndkBuild
[arm64-v8a] Compile++      : textsort_jni <= djinni_main.cpp
In file included from ../../../support-lib/jni/djinni_main.cpp:20:0:
../../../support-lib/jni/djinni_support.hpp:199:17: error: 'mutex' in namespace 'std' does not name a type
     static std::mutex       & get_mutex() { static std::mutex mtx;     return mtx; }
                 ^
../../../support-lib/jni/djinni_support.hpp: In static member function 'static djinni::static_registration<Key, T>::registration_map djinni::static_registration<Key, T>::get_all()':
../../../support-lib/jni/djinni_support.hpp:185:31: error: 'mutex' is not a member of 'std'
         const std::lock_guard<std::mutex> lock(get_mutex());`

artwyman:
Hmm, if #include <mutex> doesn't define std::mutex, that's certainly a serious problem with the STL.  Might be a bug in the beta will be fixed?  Unless they're officially de-supporting GCC, which would be pretty quick.

Jakouf:
This is in the Release Notes from NDK R11:

`Important announcements
We strongly recommend switching to Clang.
If you experience problems with Clang, file bugs here for issues specific to Clang in the NDK. For more general Clang issues, file bugs by following the instructions on this page.`

artwyman:
Yeah, I saw that, though usually that kind of strong recommendation doesn't mean "the old way will be broken tomorrow".  Ideally I'd like Djinni to be agnostic on the issue, so devs can choose whatever toolchain meets their needs.

Jakouf:
Could also be a gradle problem. In the Release notes from r12 beta is following note:


`r12 will most likely not work with Android Studio/Gradle right away. The Gradle plugin will need to be updated to match the build changes we made in ndk-build`

PR #226, opened 2016-04-24
Generation bridging header for Swift

michal-kowalczyk:
tl;dr: Introduced --objc-bridging-header argument that allows to generate bridging header Objective-C->Swift. 

In order to use Objective-C(++) in Swift, Objective-C Bridging Header must be created. Such header consists of imports of all Objective-C headers that will be used in Swift code. For support of Objective-C++ (generated code in djinni is a case) also *.mm file must be created that will import Objective-C Bridging Header.

Due to the fact that such Bridging Header content consists of lines that depends on generated content by djinni, whole Bridging Header could be generated too. In this PR you will find changes that generate Objective-C Bridging Header (and associated *.mm file for Objective-C++ support).


smarx:

@michal-kowalczyk, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

michal-kowalczyk:
CLA signed. (-:

smarx:
				
@michal-kowalczyk, thanks for signing the CLA!

example/objc/TextSort/TXSTextboxListenerImpl.swift
```diff
@@ -0,0 +1,22 @@
+//
```
artwyman: Can you put new hand-written source files under example/handwritten-src?  It's important for clarity that a new developer be able to know what kind of source they're looking at.  (I'd suggest a new directory handwritten-src/swift in this case.)

example/objc/TextSort/TXSTextboxListenerImpl.swift
```diff
@@ -0,0 +1,22 @@
...
+    
+    @objc(initWithUITextView:)
+    init (textView: UITextView) {
+        textView_ = textView
+    }
+
+    @objc func update(items: TXSItemList) {
+        textView_.text = items.items.joinWithSeparator("\n")
+    }
+}
```
artwyman: Missing newline at end of file.

example/objc/TextSort/TextSort-Bridging-Header.h
```diff
@@ -0,0 +1,7 @@
+// AUTOGENERATED FILE - DO NOT MODIFY!
```
artwyman: Generated sources should go into example/generated-src, for clarity.

example/run_djinni.sh
```diff
@@ -28,6 +28,8 @@ java_package="com.dropbox.textsort"
 
 gen_stamp="$temp_out/gen.stamp"
 
+objc_bridging_header="$base_dir/objc/TextSort/TextSort-Bridging-Header.h"
```
artwyman: I don't love the asymmetry of this option vs. the other generation options.  Also, it appears that it's generating 2 files, even though you're giving it the name of only one.  I'd suggest making the argument specify a directory for bridging-header output, calling it something like --objc-swift-bridging-out, for symmetry with the other "-out" options.

src/source/SwiftGenerator.scala
```diff
@@ -0,0 +1,59 @@
...
+  */
+
+package djinni
+
+import java.io.{File, FileWriter, BufferedWriter}
+
+import djinni.ast._
+import djinni.generatorTools._
+
+class SwiftGenerator(spec: Spec) extends Generator(spec) {
```
artwyman: This should maybe be called SwiftBridgeGenerator or something similar?  It should be clear that it doesn't actually generate Swift code, which is something which might be added in future (if someone tries to create a more Swifty interface).

artwyman:
Thanks for the submission.  I don't use Swift myself, but think this is a good inclusion in Djinni.
I've made some suggestions for clarity and functionality of the new option, and the example code.
I also think it would be valuable to add Swift header generation to the test suite, so that it's tested out in combination with more Djinni features.

4brunu:
I would really like to see this PR get merged.
@michal-kowalczyk Do you plan to finish it, or may I finish it?
Thanks

michal-kowalczyk:
@4brunu thanks for taking prompt action on this feature, I merely remembered about that. (-:

4brunu:
@artwyman Since https://github.com/dropbox/djinni/pull/301 got merged, I think this one can be closed.
@michal-kowalczyk thanks for your great work 👍

artwyman:
Supplanted by #301.

michal-kowalczyk:
@4brunu Thanks for great work! Thanks for both of you for great cooperation. I'm happy that this feature is finally done! (-:

PR #227, opened 2016-04-25
Cleaner fix for VS2015

menuet:
I added and used more macros in djinni_common for hiding VS2015 limitations, instead of having several #ifdef _MSC_VER everywhere

### DIFF
--- a/support-lib/djinni_common.hpp
+++ b/support-lib/djinni_common.hpp
@@ -16,8 +16,18 @@
 
 #pragma once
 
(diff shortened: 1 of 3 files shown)


smarx:

@menuet, it looks like you've already signed the Dropbox CLA. Thanks!

menuet:
Hi
Do you think that this commit is ok to be merged or does it need some fix ?

j4cbo:
This looks good. Thanks!

PR #240, opened 2016-06-20
Fix python cwrapper generation issue

steweg:
This patch fixes and issue, when multiple source djinni files
are using same kind of compisite object (e.g. list<...>),
which lead to creation of files with the same name more than once,
which in the end resulted in GeneralException.


smarx:

steweg:
I found out that the proposed patchset was not working as it should be

PR #241, opened 2016-06-20
Fix multiple python related issues

steweg:
- cwrapper generator to handle scenario when multiple
  IDL files are using the same compound structures (list<...> etc.)
- pycffiDynamicLibList attribute to properly expand required fields
  for proper CFFI linkage
- python code to be used with python3
- support-lib related to python to use environment variable instead
  of hardcode for getting "ffi" and "lib" attributes


smarx:

artwyman:
Looks like there are conflicts to resolve (possibly with your other PR which I just merged).

Also, can you re-run the generators for the example and test-suite so I can see the results?  I'm guessing the changes here should have some visible effect, though if not I'd encourage you to add new test cases which demonstrate the fixes you're making here.  You should be able to run "make all" from the root dir, which will regenerate everything and run all the tests to validate everything works.

Note that I don't generally recommend committing generated files to source-control in most projects, but in the Djinni repo we commit generated files to GitHub because they ease the code-review process, and helps show the impact of a change to the generator.

steweg:
Hi, I have rebased the patchset. I was trying to also regenerate the generated files, but unortunately I was not able to do so as "make all" or "make python" requires xcodebuild command, which I don't have on Debian (or any other Linux except MacOS). Can you suggest some idea how to do that?

artwyman:
Xcode is only required for the iOS portions of the build, so you should be able to run everything else, though there's no convenient make target for that.  Since you're working on changes which only affect Python you should be able to do something like this:

    make example_python
    make -C test-suite python

Unrelatedly, I started glancing through the code, and would like to hear more about the rationale behind the new environment variable.  I'm not comfortable with the support library requiring an environment-variable to be set in order to be funcitonal, since that seems very fragile.  It seems like there should at least be a fallback option if the variable is not set.  But I'd like to understand better what use case the variable is for so I can consider whether there's an alternative I might suggest.

steweg:
Hi,

make example_python requires xcodebuild (very strange dependency from my point of view): Relevant snippet from Makefile:
```
./build_py/cffi/libtextsort_py.dylib: ./build_py/example/libtextsort.py.xcodeproj/project.pbxproj
        xcodebuild -project ./build_py/example/libtextsort.py.xcodeproj \
           -target libtextsort_py \
           -configuration 'Debug' \
           ONLY_ACTIVE_ARCH=NO
        mkdir -p ./build_py/cffi
        cp example/build/Debug/libtextsort_py.dylib ./build_py/cffi/libtextsort_py.dylib

example_python2: ./build_py/cffi/libtextsort_py.dylib
        ./example/run_py_example.sh python

example_python3: ./build_py/cffi/libtextsort_py.dylib
        ./example/run_py_example.sh python3

example_python: example_python2 example_python3
```

The other command ended up in the similar situation, when in file test-suite/PyCFFI.mk, it is expected to use "-arch" and "-dynamiclib", all of them are related to building of C/C++ for OSX. These options are not recognized on linux platforms (I am using gcc version 4.9.2 (Debian 4.9.2-10))

So the testing scripts are written in a way that I am not able to test them at all (without major rewrite and as such breaking the whole point of testing). This might need to be done in by someone who understood better than I do, how to the whole testing part should work.


Now back to the environment variable. The whole problems are the lines of hardcoded code within pport-lib/py/djinni/exception.py and support-lib/py/djinni/pycffi_marshal.py files. They simply expect that there will be python library with name PyCFFIlib_cffi (in the file system this is PyCFFIlib_cffi.so). This library is actually result of command from example/run_py_example.sh :
`$python_cmd" $cffi_out/pycffi_lib_build.py $wrapper_marshal ${cffi_dep[*]} `
The problem with that is that particular command creates as SO file, but named based on "--pycffi-package-name" attribute. In your tests this attribute is always set to PyCFFIlib (see file example/run_djinni.sh), so in your particular case it works without any error. If you change that attribute, you will see that it will fail.

As removing such include/imports is not an option, to solve this problem, we could either adjust the support-lib files to use environment variable, which will point the include to the right places (as I did in current patchset) or move those two files to not be part of hardcoded support-lib and rather add them as auto-generated files (this seems to me to be much more complicated solution, but from long-term solution might be more suitable).

artwyman:
Oops, sorry for misleading you on the tests.  I forgot some of the shortcuts which were taken to finish Python support for CppCon last year, one of which was building in the most expedient way for Mac only.  That shouldn't keep you from being able to do the generation phase, though, which should run on Linux (there are other Linux users of non-Python languages who have done so).  Try these commands:

    make run_example_djinni
    cd test-suite && ./run_py_djinni.sh

That should allow us to see the impact of your changes on the code generated by the example and the test suite.  However, if you plan to contribute significant changes to Djinni I'd encourage you put some time into making sure you can run the test suite on your platform, since otherwise it'll be hard to add new tests to validate your changes, as well as check for regressions.  The test suite's python build is done with a Makefile (not xcbuild) so it you shouldn't have to change the overall structure, but you'll have to add some conditionals for arguments, replace the dylib with a .so, etc.  I don't actually know the behavior and requirements of CFFI for loading libraries on Linux, which is why we didn't take a stab at it earlier.  But since you're clearly working on that, you're in a good position to figure it out. :)

I'll respond separately to the discussion of the environment variable.

artwyman:
Okay, I understand the issue behind the env var too.  Another shortcut taken, and not caught since in the initial release we didn't have more than one use case.  The ability to customize the ffi and lib objects is something we had on our backlog, though we were thinking of it as a way to support more than one Djinni-generated module in the same process, rather than the issue you discovered.

I don't like the environment variable as a solution in particular, because it's a process-wide global state, which both requires very global knowledge for users of the app which includes a Djinni library, and also won't scale well if we do get to the point of supporting multiple Djinni-generated modules in the same Python process.  Turning these support-lib helpers into generated code would fix it, and the complexity might be limited if the generation was a simple case of reading a template file and replacing some variables.   However a simpler solution would be better for your immediate needs.  An ideal solution might be to pass around the ffi and lib variables from generated code as extra arguments, but that might also require a lot of changes.

One idea: Consider placing the ffi and lib variables into globals which generated files could pre-populate as they're loading.  Imagine if the generated python files had something like this at the top:

    from WhateverPyCFFIlib_cffi import cffi, lib # This line can use the configured name from generation
    from djinni.cffi_lib import init_cffi_lib
    init_cffi_lib(ffi, lib) # The first call to this stores the variables in globals for later use.
    from djinni.support import pycffi_marshal

Then the support lib would simply import from an alternate location:

    from cffi_lib import ffi, lib

It's still a global approach, but it's within Python not a hidden thing at the OS level, and easier to tweak later to support different options.  What do you think?

steweg:
Sorry for late reply, I agree with your last proposal, I will test it and possibly adjust the patchset

steweg:
Sorry for late reply, I am currently completely overwhelmed by other work, so I will test it as soon as possible (hopefully in Feb 2017)

steweg:
I am cancelling this PR, as in the end we end up in situation that even with proposed patch, the generated files were not suitable for our purposes. To explain why:
- the CFFI generated python script is targeting to create SO file, which needs to be linked with already existing C++ SO that provides implementations of given djinni interfaces
- the CFFI generated python script also expects that the name of CFFI wrapper will be already known during generation of djinni files

What we need is actually the exact opposite:
- We need CFFI python file to generate STATIC library, that can be bundled with our C++ SO library and being shipped easily within the same package.
- We need to define the CFFI wrapper name based on definitions that are coming during building phase of our C++library (we have about 5 different products coming from the same set of sources)

Achieving the static library is not possible to be done within python CFFI, therefore we are still using the CFFI generated python script to generate the wrapper source files, but we are not linking it by python, but rather directly g++/ld tools. To achieve the correct CFFI wrapper name, we are simply replacing the predefined string within the files with correct one using sed command

Anyway I opened another one, which is still valid for python part for any kind of usage

artwyman:
Understood.  The build/packaging part of the Python prototype was something we knew needed some more work and had some assumptions in it.  It sounds like if/when that's tackled it should be made to be more generic, generating code but then leaving the details of compiling/linking that code up to the developer, which is more similar to what we do for other languages.  Sounds like you've found a way to make it work, though, so I'm glad of that.  I'll take a look at your new PR.

Issue #213, opened 2016-03-05
Missing std::hash<enum> for map arguments


artwyman:
Details copied from discussion of issue #165.

Explanation of how to reproduce the problem from @fostah:

Here's how I edited the example project. To simplify things, I just added a new static method to sort_items:

```
sort_items = interface +c {
    static test(sortMap: map<sort_order, textbox_listener>);
    // ...
}
```

Then in sort_items_impl I just have a blank implementation of test:

```
void SortItems::test(const std::unordered_map<sort_order, std::shared_ptr<TextboxListener>> & sortMap) {

}
```

When I try to compile (make example_ios), I see the following error:

```
In file included from /Users/mfoster/Workspace/TEMP/dropbox/djinni/example/generated-src/cpp/sort_items.hpp:6:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:597:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/type_traits:964:38: error: implicit instantiation of undefined template 'std::__1::hash<textsort::sort_order>'
    : public integral_constant<bool, __is_empty(_Tp)> {};
```

Note that if I change the key to a string, everything compiles fine.

Explanation from @atwyman: This issue seems to exist when compiling using C++11 only.  Changing to c++1y in common.gypi makes the problem go away.  That plus the specifics of the error makes me realize that what's missing isn't the definition of the enum, but the definition of the specialized std::hash in the enum's header. It seems that in C++14 the STL already can hash enums by default, without help? I think that rings a bell as something I heard about in the past.

Regardless, I think the place where the definition is needed isn't necessarily in the sort_items.hpp header, but in TXSSortItems+Private.mm where the map actually needs to be created. It might make sense to generate it there, or in one of the +Private header files for ObjC++. I'll look into that.

j4cbo:
[...]

PR #251, opened 2016-07-12
Fix missing full declarations in IOS files

steweg:
This patch fixes issue when djinni defined Enum class
was used in some djinni interface, as a key attribute for map or set.
In those cases IOS generated *.mm files were not able to be compiled
without modification of djinni generated content.

In order to fix it the only easy way is to disable forward declarations
for djinni enum types, and use only full header include instead. The hard
and more optimal way would be to extend the djinni with checking if the enum
is actually used by map or set within the interface and include the header file
only then, but that seems to be major refactoring.

### DIFF
--- a/src/source/CppMarshal.scala
+++ b/src/source/CppMarshal.scala
@@ -72,11 +72,10 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
         }
       case DEnum =>
         if (d.name != exclude) {
-          if (forwardDeclareOnly) {
-            List(DeclRef(s"enum class ${typename(d.name, d.body)};", Some(spec.cppNamespace)))
-          } else {
-            List(ImportRef(include(d.name)))
-          }
+          // out enums might be used in Map or Set, which requires to be fully aware about the enum type
+          // so we cannot use forward declarations in this case, otherwise interfaces with std::unordered_map
+          // or std::unordered_set will not compile (e.g. the Objc djinni generated mm files)
+          List(ImportRef(include(d.name)))
         } else {
           List()
         }


smarx:

artwyman:
@steweg Is this intended to address issue #213, or is there a different issue behind this?

steweg:
Its very similar, but not necessary the same. I have run small tests with code like this (adjusted one of the gtests):
```
#include <unordered_map>
#include <HashEnumWorkaroundForGcc.hpp>
enum class myTestEnum;
struct myTestStruct;

TEST_F(SomeTest, test1)
{
        std::unordered_map<myTestEnum, int> myMap;
        std::unordered_map<myTestStruct, int> myMap2;
}
```

It was able to compile the TestEnum part, but not TestStruct. In IOS part it seems that there is this exact problem that requires including similar file as HasEnumWorkaroundForGcc.hpp. And for structures it requires to include the full definitions of std::hash<myTestStruct> therefore including the myTestStruct header file.

So from my point of view the long term solution should be that in case of "set" or "map", scala generator code should look into first argument (the "key") and if the key is "enum" or "record" type, include the associated header file instead of forward declarations. Unfortunately I was not able to find a way how to extend the code in such way (not a scala expert, so maybe this will be pretty easy for you to fix)

artwyman:
I assume those gtests are a part of your codebase not Djinni's.  What error are you seeing?

It would be useful if you could provide a way to reproduce this issue in the Djinni test-suite.  That would allow me to reproduce it, as well as being an important validation of any fix which gets merged.  I should have some time to look into fixing this this week, and will work on a fix for the case in #213, but if you suggest that your case is different, then I need to understand that case too in order to be sure of fixing them both.

steweg:
Just use this djinni as an example. And try to compile the generated "*.mm" files in some IOS project

```
StructData = record {
 var1 : string;
}

EnumType = enum {
 LEVEL_1;
 LEVEL_2;
 LEVEL_3;
}

MyInterface = interface +o {
 Get() : map<EnumType, binary>;
}

MyInterface2 = interface +o {
 Get() : map<StructData, binary>;
}

```

artwyman:
@steweg please take a look at #259 as an alternative, and let me know if it fixes your particular issue.

artwyman:
Supplanted by #259

PR #254, opened 2016-07-15
Cpp wide strings support

konovalov-aleks:
Added C++ wide strings support


smarx:

@konovalov-aleks, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@konovalov-aleks, thanks for signing the CLA!

support-lib/jni/djinni_support.cpp
```diff
@@ -371,6 +371,15 @@ jstring jniStringFromUTF8(JNIEnv * env, const std::string & str) {
     return res;
 }
 
+jstring jniStringFromWString(JNIEnv * env, const std::wstring & str) {
+
+    std::u16string utf16(str.begin(), str.end());;
```
mknejp: Is this portable? The C++ spec doesn't specify what character encoding `wstring` uses. There is no guarantee it's UTF-16.

support-lib/jni/djinni_support.cpp
```diff
@@ -441,6 +450,20 @@ std::string jniUTF8FromString(JNIEnv * env, const jstring jstr) {
...
+    jniExceptionCheck(env);
+
+    const auto deleter = [env, jstr] (const jchar * c) { env->ReleaseStringChars(jstr, c); };
+    std::unique_ptr<const jchar, decltype(deleter)> ptr(env->GetStringChars(jstr, nullptr),
+                                                        deleter);
+
+    const char16_t* begin = reinterpret_cast<const char16_t *>(ptr.get());
+    const char16_t* end = begin + length;
+    return std::wstring(begin, end);
+}
```
mknejp: Same as above

support-lib/jni/djinni_support.cpp
yzhong52: I guess @konovalov-aleks wants to ensure that it is using `char16_t` (std::u16string) so that it can be converted to `unsigned 16 bits` (jchar)? But `std::wstring` is using `wchar_t`, which can be either 16 bits or 32 bits depending on the compiler. This conversion `std::u16string utf16(str.begin(), str.end())` is not safe though.

mknejp: It doesn't really matter what the size of `wchar_t` is if it can be assigned to `char16_t` without error. But later down `env->NewString()` expects Unicode characters which is not guaranteed to be the native encoding of `wchar_t`/`wstring`.

j4cbo: +1. Ideally we'd have a compile-time check to see whether `wchar_t` is of size 2 or 4 and behave appropriately either way. Failing that, there should at least be a `static_assert` with a clear message (e.g. "UCS-4 builds are currently unsupported")

test-suite/handwritten-src/objc/tests/DBWcharTests.m
```diff
@@ -0,0 +1,33 @@
...
+
+- (void)tearDown
+{
+    [super tearDown];
+}
+
+- (void)test
+{
+    NSString *str1 = @"some string with unicode \u263a symbol";
+    NSString *str2 = @"another string with unicode \u263b symbol";
```
j4cbo: Please test some non-BMP characters (e.g. U+1F4A9, encodable in ObjC as `\U0001F4A9`, or most other emoji) as well

support-lib/jni/djinni_support.cpp
artwyman: I don't object to assuming UTF-16, since we already assume UTF-8 for std::string, which the standard also doesn't say as far as I know.  Assuming 16-bit wchar_t is more questionable.  If it compiles and works, I'd be okay with keeping the assumption (and documenting it) presuming nobody has the need for UCS-4 right now, but UTF-16 in 4-byte chars can work just fine.  If it doesn't work, a clear assert like @j4cbo suggests is a good idea.

artwyman:
I'm going to wait for an update to address the comments so far before I do a review.  Looks like this'll also need a merge/rebase due to conflicts.

test-suite/run_djinni.sh
```diff
@@ -73,6 +75,35 @@ fi
     --cpp-optional-template "std::experimental::optional" \
     --cpp-optional-header "<experimental/optional>" \
     --cpp-extended-record-include-prefix "../../handwritten-src/cpp/" \
+    --cpp-use-wide-strings true \
```
artwyman: Can you re-structure this and add some comments to make it clear that Djinni is being run twice, and why that is the case?  There are a few tricky details here, like the fact that you're letting only the second command fill in the list of infiles/outfiles.  That works for the test suite, though it wouldn't work if those were actually being used to detect dependency changes.

We desperately need a higher-level structure of the test suite which allows for multiple Djinni runs as separate test cases.  Until that's developed, this is a reasonable approach, and works since there's no sharing between your wchar_test.djinni and the main all.djinni file.

test-suite/run_djinni.sh
```diff
@@ -73,6 +75,35 @@ fi
...
+    \
+    --jni-out "$temp_out_relative/jni" \
+    --ident-jni-class NativeFooBar \
+    --ident-jni-file NativeFooBar \
+    \
+    --objc-out "$temp_out_relative/objc" \
```
artwyman: Why don't I see generated sources in these various output directories in this diff?  I'd expect to.
Note that we include generated output in the Djinni repo because it eases code review and makes it easier to see how a generator change affects the generated code.  It's not a best-practice for users of Djinni.

support-lib/jni/djinni_support.cpp
```diff
@@ -22,6 +22,7 @@
 #include <cstring>
 
 static_assert(sizeof(jlong) >= sizeof(void*), "must be able to fit a void* into a jlong");
+static_assert(sizeof(wchar_t) == 2 || sizeof(wchar_t) == 4, "wchar_t must be represented by UTF-16 or UTF-32 encoding");
```
artwyman: I'm curious whether in common usage a 4-byte wchar_t definitely indicates an assumption that UTF-32 is being used, as opposed to maybe UTF-16 which happens to have been copied into wider chars.  There's no enforced requirement either way, of course, any more than an 8-bit string is guaranteed to be UTF-8, rather than latin-1 or some other 8-bit charset.  I think the assumptions made here are as good as they need to be, I just wonder if they might bite someone later, and whether we'll really need a 3-way option to support 8/16/32 bit strings.

support-lib/objc/DJIMarshal+Private.h
```diff
@@ -118,6 +118,27 @@ struct String {
...
+        NSStringEncoding encoding = CFStringConvertEncodingToNSStringEncoding(kCFStringEncodingUTF32LE);
+        NSData* data = [string dataUsingEncoding:encoding];
+        return std::wstring((wchar_t*)[data bytes], [data length] / sizeof (wchar_t));
+    }
+
+    static ObjcType fromCpp(const CppType& string) {
+        assert(string.size() <= std::numeric_limits<NSUInteger>::max());
+        return [[NSString alloc] initWithBytes:string.data()
+                                        length:string.size() * sizeof(wchar_t)
+                                      encoding:NSUTF32LittleEndianStringEncoding];
```
artwyman: I don't understand why the encoding constants here are fixed to UTF32.  Shouldn't there need to be some variability between UTF16 and UTF32 as there is in Java?  Are you assuming that whcar_t is always 32-bits in Objective-C?  That assumption seems to be sound for iOS and OSX, though I'm concerned it might not be truly universal, or unchangeable.  I think the code here should at least have a static_assert to check that assumption, if you're not going to go so far as to implement the 16-bit option.

artwyman:
Thanks @konovalov-aleks.  I've done a review and made some suggestions in line comments.
I'd also love to get another look from @j4cbo before this merges, since he's behind most of the Unicode support in Djinni.

support-lib/jni/Marshal.hpp
```diff
@@ -178,6 +178,25 @@ namespace djinni
...
+
+        static CppType toCpp(JNIEnv* jniEnv, JniType j)
+        {
+            assert(j != nullptr);
+            return jniWStringFromString(jniEnv, j);
+        }
+
+        static LocalRef<JniType> fromCpp(JNIEnv* jniEnv, const CppType& c)
+        {
+            return {jniEnv, jniStringFromWString(jniEnv, c.c_str())};
```
j4cbo: The `c_str()` here shouldn't be necessary (and it will break handling of strings containing U+0000)

support-lib/objc/DJIMarshal+Private.h
j4cbo: +1 (and likewise in `toCpp` above). I'd suggest a static_assert

support-lib/objc/DJIMarshal+Private.h
```diff
@@ -118,6 +118,27 @@ struct String {
...
 
+struct WString {
+    using CppType = std::wstring;
+    using ObjcType = NSString*;
+
+    using Boxed = WString;
+
+    static CppType toCpp(ObjcType string) {
+        assert(string);
+        NSStringEncoding encoding = CFStringConvertEncodingToNSStringEncoding(kCFStringEncodingUTF32LE);
```
j4cbo: Why can't this just be `NSUTF32LittleEndianStringEncoding`?

support-lib/jni/Marshal.hpp
artwyman: Good point, though it looks like the 8-bit version above has the same issue when it calls jniStringFromUTF8().  That seems like a bug worth fixing.

support-lib/jni/djinni_support.cpp
konovalov-aleks: C++ standard has no any guarantees about wchar_t representation. But almost all modern compilers uses utf-16 or utf-32 to represent wide characters and many serious projects adopt such assumptions (for example, Python`s unicode object: https://github.com/python/cpython/blob/2.7/Objects/unicodeobject.c).

support-lib/jni/Marshal.hpp
konovalov-aleks: Bug fixed. Thank you

support-lib/objc/DJIMarshal+Private.h
artwyman: @konovalov-aleks still looking for a static_assert to check your assumption that wchar_t is 4 bytes, though it strikes me that maybe that assert should go into the generated code which actually uses wstring.  Otherwise the support-lib code would force a compile failure even if wstrings are not in use.

support-lib/jni/djinni_support.cpp
artwyman: I think it would be better if this static_assert didn't trigger if wstrings aren't in use, since they're an optional feature.  I think you could instead put a static_assert inside of the default implementation of your template functions, so it would only trigger if that function is called, and thus instantiated.

artwyman:
I made some suggestions about static_asserts in line comments.  Otherwise this looks good.

support-lib/jni/djinni_support.cpp
mknejp: If the expression inside static_assert has no dependent names it is evaluated even if not instantiated.

artwyman: Yeah, but I was assuming if you put it inside of template functions you'd assert on a type argument.  I guess what you're saying is that a static_assert(false) inside of the default implementation of the template functions below wouldn't work.  But a static_assert(wcharTypeSize == 2 || wcharTypeSize == 4) would work, right?  The checking is redundant, but the use of the template parameter will delay the evaluation.

konovalov-aleks:
I corrected the code in accordance with your comments. Please view it again

artwyman:
LGTM.  Looks like there are some conflicts which changes to master since this PR was submitted, so please resolve those with a merge or rebase, then I'll merge.

konovalov-aleks:
I resolved the conflict

artwyman:
@konovalov-aleks were you able to build and run all the examples/tests before landing this?  Are you building on Mac?  I'm assuming so, but when I pulled this I ran into issues I can't seem to get around when building the Java test suite, as well as the localhost example.  Both of those are local Java  builds running on my Mac (vs. the iOS and Android builds which seem to work fine).

I'm getting undefined symbols for basic wstring stuff (begin() and end() as shown below.  I've tried adding extra arguments for -stdlib=libc++ and -lc++.  I've tried using the clang compiler from Xcode 8 (beta) and 7.3.0.  I'm puzzled.

```
     [exec] Undefined symbols for architecture x86_64:
     [exec]   "std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> >::end() const", referenced from:
     [exec]       std::__1::basic_string<char16_t, std::__1::char_traits<char16_t>, std::__1::allocator<char16_t> > djinni::implWStringToUTF16<2>(std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> > const&) in djinni_support.cpp.o
     [exec]   "std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> >::begin() const", referenced from:
     [exec]       std::__1::basic_string<char16_t, std::__1::char_traits<char16_t>, std::__1::allocator<char16_t> > djinni::implWStringToUTF16<2>(std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> > const&) in djinni_support.cpp.o
     [exec] ld: symbol(s) not found for architecture x86_64
     [exec] clang: error: linker command failed with exit code 1 (use -v to see invocation)
     [exec] make[3]: *** [libDjinniTestNative.dylib] Error 1
     [exec] make[2]: *** [CMakeFiles/DjinniTestNative.dir/all] Error 2
     [exec] make[1]: *** [all] Error 2
```

artwyman:
I've logged issue #264 to track and discuss this.  If this is a widespread problem I might need to revert the merge.

PR #257, opened 2016-07-21
Android parcelable support

konovalov-aleks:
implemented support for android.os.parcelable


smarx:

mknejp:
Hmm we're back at the problem of stuffing the YAML files with additional content, in this case it is even mandatory for projects where the feature isn't actually used. Considering the `Parcelable` interface requires `writeToParcel` and `CREATOR` do we need the YAML entries?

test-suite/djinni/date.yaml
```diff
@@ -23,6 +23,8 @@ java:
   reference: true
   generic: true
   hash: '%s.hashCode()'
+  writeToParcel: 'out.writeString(%s.toString())'
+  readFromParcel: 'new %s(in.readString())'
```
mknejp: This is using a deprecated constructor. Wouldn't it be more efficient to use `getTime()` and `Date(long)` instead?

test-suite/djinni/duration.yaml
```diff
@@ -26,6 +26,8 @@ java:
   reference: true
   generic: false
   hash: '%s.hashCode()'
+  writeToParcel: 'out.writeString(%s.toString())'
+  readFromParcel: '%s.parse(in.readString())'
```
mknejp: Same as with `Date` above. I think a better alternative is using `toNanos()` and `ofNanos(long)` instead of parsing strings.

test-suite/handwritten-src/java/android/os/Parcel.java
```diff
@@ -0,0 +1,155 @@
+package android.os;
```
artwyman: Did you write this from scratch, or copy it from somewhere?  It needs a license, and/or a comment explaining it's purpose, particularly pointing out that this is a mock replacement for Android's implementation.

test-suite/handwritten-src/java/android/os/Parcel.java
```diff
@@ -0,0 +1,155 @@
...
+
+public final class Parcel {
+
+    private FileOutputStream mOutFile;
+    private FileInputStream mInFile;
+    private ObjectOutputStream mOut;
+    private ObjectInputStream mIn;
+
+    public Parcel() {
+        final String fname = "parcel.tmp";
```
artwyman: This will create this temp-file somewhere (path unspecified) and never clean it up?  I'd prefer to see the test suite explicitly create the file in a known location and clean it up after the tests are done.

src/source/JavaGenerator.scala
```diff
@@ -357,6 +357,131 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
         }
         w.wl
 
+        if (spec.javaImplementAndroidOsParcelable) {
+          // CREATOR
```
artwyman: Can you pull out these various long chunks of code into helper functions each with a clear purpose?

artwyman:
I've made some suggestions in line comments.

As a high-level comment I wonder if this command-line argument approach is the best, or if it would be better to let Parcelable be specified on individual records using a deriving clause (similar to "deriving ord").  That would provide conditionality in a way which is clear in the code where parcelability is desired, as well as making it optional and thus not forcing it to be implemented on external types in yaml if it won't be used.

Maybe for maximal flexibility you'd need both a deriving clause to say that parcelable is desirable on Android, and a cmd-line arg indicating that the Java code is for Android.  That would only be necessary if the same IDL needs to support Android Java and non-Android Java generation in separate runs.

konovalov-aleks:
Hi @artwyman, thanks for your feedback!
I've implemented your suggestions. Please review my code

test-suite/djinni/primtypes.djinni
```diff
@@ -14,4 +14,4 @@ assorted_primitives = record {
     o_sixtyfour: optional<i64>;
     o_fthirtytwo: optional<f32>;
     o_fsixtyfour: optional<f64>;
-} deriving (eq)
+} deriving (eq,parcelable)
```
artwyman: Nit: Needs newline at end of file.

src/source/meta.scala
```diff
@@ -57,7 +57,9 @@ object MExtern {
     boxed: String, // Java typename used if boxing is required, must be an object.
     reference: Boolean, // True if the unboxed type is an object reference and qualifies for any kind of "nonnull" annotation in Java. Only used for "record" types.
     generic: Boolean, // Set to false to exclude type arguments from the Java class. This is should be true by default. Useful if template arguments are only used in C++.
-    hash: String // A well-formed expression to get the hash value. Must be a format string with a single "%s" placeholder. Only used for "record" types types with "eq" deriving when needed.
+    hash: String, // A well-formed expression to get the hash value. Must be a format string with a single "%s" placeholder. Only used for "record" types types with "eq" deriving when needed.
+    writeToParcel: String, // A well-formed expression to write value into android.os.Parcel. Must be a format string with a single "%s" placeholder. Only used for "record" types types
```
artwyman: I'm not familiar enough with the external type / YAML support.  Are these new fields considered optional? What will happen if they're missing?  My main concern is that this new feature shouldn't break any existing users with yaml files who aren't depending on this new feature.  It's fine for users to have to add these new fields as soon as they want to derive parcelable.

test-suite/generated-src/cpp/enum_usage_record.cpp
```diff
@@ -0,0 +1,9 @@
+// AUTOGENERATED FILE - DO NOT MODIFY!
```
artwyman: These don't seem necessary to generate in C++ as a result of the addition of deriving(parcelable) which only affects Java.  There's probably some logic in the C++ generator which will need to be tweaked to look for specific deriving flags, not just for any deriving flag.

4brunu:
This feature would be great!
@konovalov-aleks any news on this?
Thanks

4brunu:
@konovalov-aleks This feature is great!
Since this PR is inactive for some time, I pick the changes here and make the changes requested by @artwyman.
Hope this is ok to you. 
If you want, you can pick my PR and continue the work here :)
Thanks

artwyman:
Replaced by #308

PR #259, opened 2016-07-28
Fix enum compile issue in C++11

artwyman:
Fix for #213 and alternative to #251.

The fix involved introducing a DBMyEnum+Private.h generated file, in order to include my_enum.hpp only where necessary (in the ObjC++ private sources).  I needed to generate the new files to maintain the assumption (in the generators, and yams definitions) that each type needs only one header for marshaling.

In order to validate this properly, I converted the test-suite to use C++11 rather than C++14, which is a good idea since Djinni is intended to work with C++11.  This conversion required importing a reference definition of std::experimental::optional.

I can merge this myself, but am posting here for comment.  I'm particularly hoping that @steweg will try this out to confirm that it fixes his issue, and that @xianwen and @mknejp will take a look and comment.


artwyman:
A few notes:
- I think that because the new generated files are headers, this won't actually break anyone's compilation, though those headers may not show up in people's projects.  That was already truew of a lot of headers in the test suite, as you can see me fixing here.
- I added some additional usages of enums in the test suite, though even without that the existing test-suite triggered the bug when compiled with C++11 due to the check_enum_map method in test_helpers.

src/source/ObjcppMarshal.scala
```diff
@@ -34,9 +34,7 @@ class ObjcppMarshal(spec: Spec) extends Marshal(spec) {
     case o: MOpaque =>
       List(ImportRef(q(spec.objcBaseLibIncludePrefix + "DJIMarshal+Private.h")))
     case d: MDef => d.defType match {
-      case DEnum =>
-        List(ImportRef(q(spec.objcBaseLibIncludePrefix + "DJIMarshal+Private.h")))
```
mknejp: Instead of creating a new file couldn't we just add the C++ header to this list?

artwyman: I couldn't just add the C++ header to the list in references() because of the need to return only a single header file name in include() below.  That's called from yaml generation, in a spot which seems to expect only a single header file, not a list of them.

steweg:
Hi, I have tested it, and it works as expected. Thanks

src/source/ObjcppMarshal.scala
```diff
@@ -49,7 +47,6 @@ class ObjcppMarshal(spec: Spec) extends Marshal(spec) {
 
   def include(m: Meta) = m match {
     case d: MDef => d.defType match {
```
xianwen: Looks like this `match` here is no longer needed, you can simply return the result here.

Issue #245, opened 2016-06-30
Possible to make generated Java code package-private?


avf:
Hi,

I might be doing this wrong, but I'm writing an Android library which uses djinni and I'd like to keep the generated code out of my library's public API. As far as I can tell, the only way to do this would be to have the djinni `--java-package` be the same package as my library and then remove the `public` access modifier from the djinni generated sources. There doesn't seem to be a way yet to do this automatically.
I have a feeling I'm approaching this the wrong way. Is there a better way to achieve this?

artwyman:
There's no existing feature for this.  The approach you suggest would work, and it wouldn't be too difficult to add a Djinni command-line argument to specify the visibility of generated Java classes (but it would take changes to the Djinni implementation, obviously).

Another approach I've taken to similar problems in the past (in pre-Djinni days) was to use class-file modifier to change visibility as a post-build step.

PR #260, opened 2016-07-28
Added command line option for specifying the access modifier for generated java classes

avf:
This pull request adds the feature requested in issue #245. Please be gentle, this is my first time writing Scala code. Let me know if I can make any improvements before merging.

### DIFF
--- a/src/source/generator.scala
+++ b/src/source/generator.scala
@@ -141,14 +141,12 @@ package object generatorTools {
 
   object JavaAccessModifier extends Enumeration {
     val Public = Value("public")
-    val Protected = Value("protected")
     val Package = Value("package")
 
     def getCodeGenerationString(javaAccessModifier: JavaAccessModifier.Value): String = {
       javaAccessModifier match {
         case Public => "public "
-        case Protected => "protected "
-        case Package => ""
+        case Package => "/*package*/ "
       }
     }
 
--- a/src/source/generator.scala
+++ b/src/source/generator.scala
@@ -30,6 +30,7 @@ package object generatorTools {
   case class Spec(
                    javaOutFolder: Option[File],
                    javaPackage: Option[String],
+                   javaClassAccessModifier: JavaAccessModifier.Value,
                    javaIdentStyle: JavaIdentStyle,
                    javaCppException: Option[String],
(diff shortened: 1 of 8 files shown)


smarx:

@avf, it looks like you've already signed the Dropbox CLA. Thanks!

src/source/generator.scala
```diff
@@ -137,6 +138,22 @@ package object generatorTools {
     }
   }
 
+  object JavaAccessModifier extends Enumeration {
+    val Public = Value("public")
+    val Protected = Value("protected")
```
artwyman: I don't think that "protected" is a valid value for top-level classes.  It would be meaningless, so I think it's disallowed.

http://stackoverflow.com/questions/3869556/why-a-class-cannot-be-defined-as-protected

I think public and package-private are the only values this option needs to support.

src/source/generator.scala
```diff
@@ -137,6 +138,22 @@ package object generatorTools {
...
+    val Package = Value("package")
+
+    def getCodeGenerationString(javaAccessModifier: JavaAccessModifier.Value): String = {
+      javaAccessModifier match {
+        case Public => "public "
+        case Protected => "protected "
+        case Package => ""
```
artwyman: In other cases where package-private is used by Djinni (e.g. in constant fields) the generator outputs a comment /*package*/ in place of the accessibility.

artwyman:
I'm not really a guru on Scala style either, but this looks pretty good to me.  I made some suggestions in line comments.

I'd like to see the new argument exercised in the test-suite or the example, so we have some validation that it works, and can see the effect on the generated code.  Ideally, I'd suggest changing one of the two to use a non-default value (package), while the other uses the default value, so both can be seen in action by someone examining code in the repo.  A quick test suggests that the example will still compile unchanged with package-private, since it's all one package.  The test-suite doesn't, since tests are in a separate package.

avf:
Hi @artwyman, thanks for your feedback!
I've implemented your suggestions and removed the `protected` option, added `/*package*/` to the generator and updated the sample to use the `package` option.

artwyman:
LGTM.  Thanks @avf.

Issue #264, opened 2016-08-04
Undefined symbols building for Java on Mac


artwyman:
This seems to be fallout since merging PR #254, which added wstring support.  I'm not sure yet whether this is a problem specific to my build environment or a general problem.  I'm interested to know else is seeing this, and if anyone can suggest a fix.

I ran into issues I can't seem to get around when building the Java test suite, as well as the localhost example.  Both of those are local Java  builds running on my Mac (vs. the iOS and Android builds which seem to work fine).  I'm getting undefined symbols for basic wstring stuff (begin() and end() as shown below.  I've tried adding extra arguments for -stdlib=libc++ and -lc++.  I've tried using the clang compiler from Xcode 8 (beta) and 7.3.0.  I'm puzzled.

```
     [exec] Undefined symbols for architecture x86_64:
     [exec]   "std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> >::end() const", referenced from:
     [exec]       std::__1::basic_string<char16_t, std::__1::char_traits<char16_t>, std::__1::allocator<char16_t> > djinni::implWStringToUTF16<2>(std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> > const&) in djinni_support.cpp.o
     [exec]   "std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> >::begin() const", referenced from:
     [exec]       std::__1::basic_string<char16_t, std::__1::char_traits<char16_t>, std::__1::allocator<char16_t> > djinni::implWStringToUTF16<2>(std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t> > const&) in djinni_support.cpp.o
     [exec] ld: symbol(s) not found for architecture x86_64
     [exec] clang: error: linker command failed with exit code 1 (use -v to see invocation)
     [exec] make[3]: *** [libDjinniTestNative.dylib] Error 1
     [exec] make[2]: *** [CMakeFiles/DjinniTestNative.dir/all] Error 2
     [exec] make[1]: *** [all] Error 2
```

konovalov-aleks:
[...]

PR #265, opened 2016-08-04
Build java tests on mac workaround

konovalov-aleks:
The workaround for a bug #264 

This error is similar to a bug in the compiler. Any calls to methods of the class std::wstring in the method implWStringToUTF16<2> led to the link error. If you call methods from other places, the error did not occur. I made to implWStringToUTF16 took iterators, thus removing all calls to methods of the class std::wstring from it. This implementation works on mac and linux


smarx:

artwyman:
Thanks for diagnosing this!  I never thought moving the calls would make a difference.

PR #267, opened 2016-08-04
fix #262

### LINKED ISSUE #262: Duplicate header includes for interfaces which reference themselves
When an interface references itself as an argument, the generated JNI/ObjC++ file for that interface ends up including its own header twice (once as a "self header", and once as an interface headers).  Examples:

https://github.com/dropbox/djinni/blob/master/test-suite/generated-src/jni/NativeClientInterface.cpp
https://github.com/dropbox/djinni/blob/master/test-suite/generated-src/objc/DBClientInterface%2BPrivate.mm

This isn't dangerous, since all generated headers are safe to include more than once.  It's just a bit messy, and theoretically has a tiny compile performance cost.

It should be easy to fix.  In the JNI case, the self header is simply always generated separately from the interface headers, with no deduping.  IN the ObjC++ case the self header gets a "!" prefix to force it to be first, which inhibits the normal deduping.

konovalov-aleks:
Fixed #262 bug


smarx:

artwyman:
Looks good, thanks!

PR #282, opened 2016-10-05
Generating constexpr for constants when possible

itsff:
I've noticed that the constants in generated C++ code are always initialized in the cpp file, even for basic types like bool and numerics.

In this patch I've modified the C++ code generator to determine if a value can be initialized in the header and, if so, place it there as a constexpr. This only happens for bools, numerics, enums and optionals thereof. If any other kind of constant is present, its value will get set in the implementation file as before.

Please refer to the "generated-src" from the test suite which is part of this commit.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -91,11 +91,34 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
   }
 
   def generateHppConstants(w: IndentWriter, consts: Seq[Const]) = {
+    var needsCppGen = false
+
     for (c <- consts) {
+      var isConstExpr = true
+      val constValue = c.value match {
+        case l: Long => " = { " + l.toString + " };"
+        case d: Double if marshal.fieldType(c.ty) == "float" => " = { " + d.toString + "f };"
+        case d: Double => " = { " + d.toString + " };"
+        case b: Boolean => if (b) " = { true };" else " = { false };"
+        case e: EnumValue => " = " + marshal.typename(c.ty) + "::" + idCpp.enum(e.ty.name + "_" + e.name) + ";"
+        case _ => {
+          needsCppGen = true;
+          isConstExpr = false;
+          ";"
+          }
+      }
+
       w.wl
       writeDoc(w, c.doc)
-      w.wl(s"static ${marshal.fieldType(c.ty)} const ${idCpp.const(c.ident)};")
+      w.w(s"static ${if (isConstExpr) "constexpr" else "" } const ${marshal.fieldType(c.ty)} ${idCpp.const(c.ident)}${constValue}")
     }
+
+    if (!consts.isEmpty) {
+      w.wl
+      w.wl
+    }
+
+    needsCppGen
   }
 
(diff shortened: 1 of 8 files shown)


smarx:

@itsff, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@itsff, thanks for signing the CLA!

mknejp:
Nice idea, however the tests are only half-way there. To make sure everything links properly the header-only constants need to be ODR-used in at least two translation units. Furthermore I wouldn't count on `optional<>` being a literal in all implementations considering Djinni is C++11 only and users have to supply their own optional type.

mknejp:
`optional` is C++17, which technically isn't out yet.

src/source/CppGenerator.scala
```diff
@@ -91,11 +91,34 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
...
+      val constValue = c.value match {
+        case l: Long => " = { " + l.toString + " };"
+        case d: Double if marshal.fieldType(c.ty) == "float" => " = { " + d.toString + "f };"
+        case d: Double => " = { " + d.toString + " };"
+        case b: Boolean => if (b) " = { true };" else " = { false };"
+        case e: EnumValue => " = " + marshal.typename(c.ty) + "::" + idCpp.enum(e.ty.name + "_" + e.name) + ";"
+        case _ => {
+          needsCppGen = true;
+          isConstExpr = false;
+          ";"
```
artwyman: I'd suggest making constValue empty here, and applying the semi-colon in the final statement below.  The semicolon isn't really part of the const value.

src/source/CppGenerator.scala
```diff
@@ -127,10 +150,20 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
 
     val skipFirst = SkipFirst()
     for (c <- consts) {
-      skipFirst{ w.wl }
-      w.w(s"${marshal.fieldType(c.ty)} const $selfName::${idCpp.const(c.ident)} = ")
-      writeCppConst(w, c.ty, c.value)
-      w.wl(";")
+      val isConstExpr = c.value match {
```
artwyman: It would be nice to unify the cases for this match with the ones above in the header file.  One option would be to make a helper function which returns the constValue used in the header file, then have isConstExpr here simply test whether that value is non-empty.

src/source/CppGenerator.scala
```diff
@@ -91,11 +91,34 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
   }
 
   def generateHppConstants(w: IndentWriter, consts: Seq[Const]) = {
+    var needsCppGen = false
+
     for (c <- consts) {
+      var isConstExpr = true
+      val constValue = c.value match {
```
artwyman: It might be clearer to match on c.ty instead.  I'm not set on that, since this does make the code simpler, but it took me a while to figure out that this distinction is the reason that optionals are being treated as constexpr-eligible.  It's not an explicit case statement anywhere, but simply implicit in the fact that the constant value for an optional<T> is of type T.

PR #287, opened 2016-11-07
Adding the ability to declare properties in interfaces.

BrassLion:
This pull request allows interfaces to define properties with Djinni-supported data types that are implemented as C++ variables and accessed from Java, through auto-generated getter and setter methods, and Obj-C, through auto-generated `@property` methods.

Properties are defined in IDL files identically to Djinni record elements. Platform-specific getters and setters are auto-generated. Developers then must implement the generated C++ getters and setters and the backing variable. Properties may be preceded with a `readonly` keyword, in which case only the platform-specific getter is generated.

This pull request is mostly syntactic sugar and can be implemented by manually defining the getter and setter methods in the IDL file, but provides a more platform-native way of accessing member variables, particularly through the use of `@property` in Obj-C.


smarx:

@BrassLion, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@BrassLion, thanks for signing the CLA!

src/source/CppGenerator.scala
```diff
@@ -297,6 +300,16 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
             w.wl(s"virtual $ret ${idCpp.method(m.ident)}${params.mkString("(", ", ", ")")}$constFlag = 0;")
           }
         }
+        // Properties
```
artwyman: It would be nice if this shared more code with the method-generating codepaths, since that will help to keep the code simple and make it likely to all be updated in sync with future changes.  You should be able to create method definitions from the properties, then use the normal method-generation code.

One option to consider (though it might be too aggressive) would be to implement property generation in most languages as a transformation on the AST, by generating a new AST with methods in place of properties, then running the normal generator on them.  That could be a suitable approach for all generators which don't have special support for properties (i.e. ObjC with its @property declaration).

BrassLion:
I'll have a look into making these changes when I next get a chance. In regards to your suggestion of transforming the AST to another without properties, what way of achieving this in Scala would you be envisioning? My Scala knowledge is fairly basic, so any pointers would be appreciated!

artwyman:
I don't have specifics in mind, really.  It's more of a theoretical suggestion.  By AST I think I just mean the list of TypeDecl objects which is the `allIdl` variable in Main.

choiip:
This is a remarkable feature. The generated API for objective-c looks better. @BrassLion, any update on the changes? It is good the hear that this nice feature could be included in Djinni.

BrassLion:
I'm afraid I haven't had any time to return to these changes and I don't see myself getting back to this in the near future. Hopefully someone will be able to continue this feature or use it as inspiration. Apologies!

PR #288, opened 2016-11-10
Optionally generate interfaces in Java when possible

csimmons0:
This pull request is a continuation of #47. I've made the new behavior optional and off by default. If I understand correctly, there was concern that djinni might not properly handle a Java class that implemented multiple interfaces generated by djinni, but I tested that theory in my Android app, by having some c++ code call both bar() and baz() on the one Java object that implemented those interfaces, defined below, and I didn't have any problems.

```
hello_world = interface +c {
    static create(): hello_world;
    foo(bar_delegate: hello_bar, baz_delegate: hello_baz): string;
}

hello_bar = interface +j {
    bar(): string;
}

hello_baz = interface +j {
    baz(): string;
}
```


smarx:

@csimmons0, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@csimmons0, thanks for signing the CLA!

src/source/JavaGenerator.scala
```diff
@@ -140,7 +140,10 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
       writeDoc(w, doc)
 
       javaAnnotationHeader.foreach(w.wl)
-      w.w(s"${javaClassAccessModifierString}abstract class $javaClass$typeParamList").braced {
+      val isAbstractClass = (spec.javaGenerateInterfaces == false) || i.ext.cpp || i.methods.exists(_.static) || i.consts.size > 0
```
artwyman: This could use a comment explaining the reason for each of the clauses, and why an interface can't support them.

Also, a Java interface can have static final member variables, so you should be able to generate an interface in that case.  In fact, static and final are implied for member variables in interfaces, so the generator should probably leave them off to avoid warnings.

mknejp:
I just checked the implementation and we already forbid static methods on +j and +o interfaces (https://github.com/dropbox/djinni/blob/master/src/source/resolver.scala#L268), so this change doesn't have to be made optional or depend on the interface's contents. The generated type can always be an `interfacae`.

artwyman:
Good point, @mknejp.  I think there are two definitions of "optional" at work here, so let me see if we agree on the details:

- Regarding which Djinni interfaces can safely be represented as Java interfaces, I think you're arguing that's true for any interface which is +j but not +c.  Such an interface can have constants but not static methods, nor an inner class for the proxy, which is consistent with what Java interfaces support in Java versions <8.  I'd still call this "optional" though with a much simpler test for compatibility.

- Regarding command-line optionality, that's intended to avoid breaking any existing code which subclasses the Djinni interface using "extends" rather than "implements".  If we generate Java interfaces only for Djinni interfaces which are "+j but not +c" then I think there can still be breakages, since there will be existing code which extends the class in Java.

I think that leaves us still wanting a command-line option, and still with an approach of "generate interfaces when possible", though the logic for "when possible" can become simpler if the generator properly supports generating constants on Java interfaces.

Are you suggesting something different?

mknejp:
> Regarding which Djinni interfaces can safely be represented as Java interfaces, I think you're arguing that's true for any interface which is +j but not +c. Such an interface can have constants but not static methods, nor an inner class for the proxy, which is consistent with what Java interfaces support in Java versions <8. I'd still call this "optional" though with a much simpler test for compatibility.

I don't think this differentiation matters. Because +j +c interfaces are +j they cannot have static methods, so even the +j +c types can be an `interface` in Java.

>Regarding command-line optionality, that's intended to avoid breaking any existing code which subclasses the Djinni interface using "extends" rather than "implements". If we generate Java interfaces only for Djinni interfaces which are "+j but not +c" then I think there can still be breakages, since there will be existing code which extends the class in Java.

True. It's a shame Java uses different syntax for this.

> Are you suggesting something different?

I'm pretty sure we can always generate `interface` for anything containing the +j flag because that implicitly disables static methods, and that's the only thing holding us back (and isn't allowed in +j for symmetry with +o).

artwyman:
> I don't think this differentiation matters. Because +j +c interfaces are +j they cannot have static methods, so even the +j +c types can be an interface in Java.

I had forgotten that the rule was "+c only" not "+c at all".  However any +c interface will have the static inner class for the CppProxy, which can't exist on a Java interface.  That can (and probably should) be moved out into a separate class, but I think that's beyond the scope of what the author was intending to take on in this PR.

mknejp:
>  However any +c interface will have the static inner class for the CppProxy, which can't exist on a Java interface.

What makes you think that?

artwyman:
> What makes you think that?

An unfounded assumption, apparently  Now that I've researched it and tried it I see that's incorrect.  I'd always assumed you couldn't have classes in interfaces, and that figured prominently in the discussion on the issues which lead to this PR.  Apparently I was wrong; sorry for the confusion.

You can have a class inside an interface, but you can't make it private.  So we could allow +j +c interfaces to be Java interfaces, so long as we're okay with the CppProxy class being public, which I think is a reasonable trade-off to make.  We still couldn't produce Java interfaces for +c interfaces with static methods as-is, though it looks like there's a reasonable alternative I can propose:

Java interfaces can't have static methods (or apparently native methods, which Djinni's static methods also are) in Java <8, but static inner classes within interfaces can.  Given the interface generation option is going to be enabled by a command-line switch, and will require code changes anyway, we could have it always generate interfaces, and move the static methods into an inner class called Statics or somesuch.  There could also be another cmdline switch to generate the Java 8 version without the intermediate class.

Thoughts?

mknejp:
Just make the CppProxy constructors private and that should do.

Static methods aren't allowed for +j not because of issues with Java but because they don't work in Objective-C so this limitation exists solely for symmetry with +o.

artwyman:
Yeah, that makes it safe enough.  I don't like the "messiness" of the extra public class, but don't think it's a blocker, as I said.

src/source/JavaGenerator.scala
csimmons0: > Just make the CppProxy constructors private and that should do. (mknejp)

I believe that the constructors he's referring to are already private. Does this mean that the "i.ext.cpp" check can be removed?

> Also, a Java interface can have static final member variables . . . (artwyman)

Are you saying that the "i.consts.size > 0" check can be removed?

csimmons0:
@mknejp @artwyman 
At first, I didn't understand what you two were discussing, but I believe that I do understand it now that I've attempted to generate Java interfaces in the +c case. 

> So we could allow +j +c interfaces to be Java interfaces, so long as we're okay with the CppProxy class being public, which I think is a reasonable trade-off to make. (artwyman)

From my perspective, this isn't a great trade-off. I, too, dislike the messiness of making the CppProxy class public, and all my Djinni interfaces are either +c or +j +o. What's the use case for +j +c?

mknejp:
+j interfaces cannot have static methods, so the whole distinction logic is unnecessary. The only criteria on whether or not to create interfaces is `spec.javaGenerateInterfaces`.

src/source/JavaGenerator.scala
```diff
@@ -97,7 +97,10 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
       writeDoc(w, c.doc)
       javaAnnotationHeader.foreach(w.wl)
       marshal.nullityAnnotation(c.ty).foreach(w.wl)
-      w.w(s"public static final ${marshal.fieldType(c.ty)} ${idJava.const(c.ident)} = ")
+
+      // If the constants are part of a Java interface, omit the "static" and "final" specifiers.
+      val staticFinalString = if (forJavaInterface) "" else "static final "
+      w.w(s"public ${staticFinalString}${marshal.fieldType(c.ty)} ${idJava.const(c.ident)} = ")
```
mknejp: This is entirely unnecessary. Just leave it at `public static final`.

csimmons0:
> +j interfaces cannot have static methods, so the whole distinction logic is unnecessary. 

This code also runs to generate the Java bindings for +c interfaces. When I remove the check for static methods and run the test code, interfaces such as this:
```
cpp_exception = interface +c {
    throw_an_exception(): i32;
    static get(): cpp_exception;
}
```
produce errors such as this:
```
    [javac] /Users/christophersimmons/Documents/repos/vitalsoftware/deps/djinni/test-suite/generated-src/java/com/dropbox/djinni/test/CppException.java:14: error: static interface methods are not supported in -source 1.7
    [javac]     public static native CppException get();
```

src/source/JavaGenerator.scala
csimmons0: Andrew suggested omitting "static" and "final" to avoid warnings, and Android Studio does give me warnings when I look at an interface that includes "static," "final," or "public" in declarations of constants.

passsy:
This is very useful for `Callback` definitions. What's missing here for a merge?

artwyman:
Right now I think it's waiting for me to have time to re-build context enough to understand and review this again.  I don't think I realized there were new changes here to review, but I also haven't had time to be very responsive recently.  Will post about next steps when I can get caught up.

artwyman:
So, at risk of re-opening a bikeshed, I want to have a brief discussion about where we landed on when Java-interfaces are generated.  @csimmons0 and @mknejp I wonder what you think about the "variability" of the current solution, which will generate Java-interfaces anytime a Djinni-interface doesn't have static methods.  It's a bit weird that adding or removing a method will have the side-effect of changing the output in incompatible ways (`extends` vs. `implements`).

The alternative I'd consider is to generate Java-classes for Djinni-interfaces which are "+c only" (i.e. not also +j).  If my recollection of the options is correct, that identifies an interface which is allowed to have static methods, whether it has any right now or not, and is a decision which won't vary as easily.  The tradeoff for doing that is generating Java-interfaces slightly less often.  However I'd argue that the value of this feature is for +j or +j+c interfaces, where it allows Java classes to implement more than one of them.  On a +c interface, I don't see how generating a Java-interface is better than a Java-class.

To be clear, I think this is a discussion of small tweaks, and getting the overall feature is more important, so if we don't reach a clear consensus on this relatively quickly I'm fine with merging the current behavior (after the code-review is approved).

4brunu:
@artwyman I think I prefer the last alternative of generating always interfaces, using static methods in interface. This way be behaviour will be more coherent.
Android Studio 3.0 and the gradle plugin 3.0 went stable yesterday, so this isn't a blocking point anymore.

artwyman:
I'd be in favor of that if @csimmons0 or @4brunu wants to propose an alternative PR.

4brunu:
I think this PR only need some tweaks, do you want to do it @csimmons0?

csimmons0:
I'll see if I can get it done today.

csimmons0:
I've implemented generating interfaces regardless of the presence of static methods, but I'm afraid it doesn't work. These static methods are also native methods, which, as I understand it, aren't allowed in interfaces because native methods aren't allowed to be abstract. I run the tests and get lots of build errors such as this:

```
    [javac] /Users/csimmons/Documents/repos/onedoclily/client/deps/djinni/test-suite/generated-src/java/com/dropbox/djinni/test/WcharTestHelpers.java:12: error: modifier native not allowed here
    [javac]     public static native WcharTestRec getRecord();
```
Referring to code such as this:
```
// This file generated by Djinni from wchar_test.djinni

package com.dropbox.djinni.test;

import java.util.concurrent.atomic.AtomicBoolean;
import javax.annotation.CheckForNull;
import javax.annotation.Nonnull;

public interface WcharTestHelpers {
    @Nonnull
    public static native WcharTestRec getRecord();
    . . .
```
Is there a workaround aside from generating interfaces conditionally, as in the original request?

4brunu:
@csimmons0 could you create another branch and commit your changes there so I can make some tests?

csimmons0:
@4brunu https://github.com/csimmons0/djinni/tree/java-interfaces-optional-test

4brunu:
Thanks @csimmons0, tomorrow I will try to take a look a it, because here it's late :)

csimmons0:
@artwyman Regarding the other option that was being considered, where we generate abstract classes for "+c only" interfaces, I'm not seeing the benefit of that option over the original method. I've thought through the ways the interface, as defined in the djinni file, may change, and I haven't found any way that the change to the interface would break code. Code would break if a Java class implemented the interface and that interface changed from a Java interface to a Java abstract class or vice-versa.

If the interface changed from a Java interface to a Java abstract class, the new version of the Djinni interface must include a static method. Therefore the new Djinni interface must be +c only. The Java class that implemented the interface is now broken due to the change from Java interface to Java abstract class, but it's also broken because Java implementations of +c only interfaces aren't supported. Furthermore, the alternative implementation of this change, which looks for +c only interfaces, would have the exact same effect.

If the interface changed from a Java abstract class to a Java interface, the old version of the Djinni interface must have had a static method. Therefore its old interface must have been +c only. Therefore there was no preexisting Java implementation to break.

Also, if something starts +c only, without static methods, and then changes to +c +j,  and we were to implement this change the other way, wouldn't some generated code change, where it wouldn't change using the original method?

csimmons0:
I should mention that I have git treat Djinni-generated files as binaries, which means that changes from interfaces to abstract classes don't show up in diffs. If you were concerned about code churn in generated files, I can see the benefit to the other way of doing things.

4brunu:
@csimmons0 I tried to found a way to solve this but without success 😞

artwyman:
Static methods are not abstract (even on interfaces) so I don't think there's any fundamental reason you couldn't have `static native` methods on an interface.  I'd believe that the Java 8 designers may not have thought about that case, though.

It looks like you could solve the problem with an extra level of indirection, though, since interfaces can have static inner classes (like `CppProxy`), which can have static native methods.  I just tried this test and IntelliJ seems to accept it:

```
public interface TestInterface {
    // static native int nat(); // This won't compile
    static int nat() {
        return Inner.nat();
    };

    static class Inner {
        Inner() {}
        static native int nat();
    }
}
```

artwyman:
Comment on the discussion of "+c only" vs. "has static methods" as the trigger (maybe moot if we switch to Java 8 and can make static native methods work):

That's an excellent point that "+c only" interfaces are generally meaningless to extend in Java, and thus the likelyhood of anyone's code breaking when they add a static method is very low.  In theory, though, people might make Java-only subclasses of the method which they don't need to pass through to C++.  I'm not sure I'd prioritize this use case highly given Djinni's purpose, but it could be valid in cases where all your users of your interface are in Java, but you want to have two different implementations in Java vs. C++.  The Java implementation might even be a mock for testing.

I'm not worried about generated files causing churn in source-control or reviews, since in general I don't recommend putting generated files in source-control at all.  In general, I think it's always fine if Djinni changes generated output in response to an IDL change (or a Djinni change), so long as the output is compatible with what the user expects. Doing otherwise would but a lot more limitations on the ability to change Djinni's behavior.

csimmons0:
I've completed the suggested changes.

test-suite/generated-src/java/com/dropbox/djinni/test/Conflict.java
```diff
@@ -11,9 +11,9 @@
  * Test for conflict of method name with an interface name.
  * See the comments about scopeSymbols in CppMarshal.scala for more info.
  */
-public abstract class Conflict {
+public interface Conflict {
 
-    private static final class CppProxy extends Conflict
+    static final class CppProxy implements Conflict
```
4brunu: Shouldn't this be private?

test-suite/generated-src/java/com/dropbox/djinni/test/ConstantsInterface.java
```diff
@@ -8,62 +8,62 @@
...
 
-    public static final byte I8_CONSTANT = 1;
+    byte I8_CONSTANT = 1;
 
-    public static final short I16_CONSTANT = 2;
+    short I16_CONSTANT = 2;
 
     /** i32_constant has documentation. */
-    public static final int I32_CONSTANT = 3;
+    int I32_CONSTANT = 3;
```
4brunu: shouldn't this continue to be public static final?

test-suite/generated-src/jni/NativeCppException.cpp
```diff
@@ -29,7 +29,7 @@ CJNIEXPORT jint JNICALL Java_com_dropbox_djinni_test_CppException_00024CppProxy_
     } JNI_TRANSLATE_EXCEPTIONS_RETURN(jniEnv, 0 /* value doesn't matter */)
 }
 
-CJNIEXPORT jobject JNICALL Java_com_dropbox_djinni_test_CppException_get(JNIEnv* jniEnv, jobject /*this*/)
+CJNIEXPORT jobject JNICALL Java_com_dropbox_djinni_test_CppException_00024CppProxy_get(JNIEnv* jniEnv, jobject /*this*/)
```
4brunu: Could you please explain the 00024CppProxy?

4brunu:
@csimmons0 This is awesome! Great work! 
I just annotated some missing public, private, static and final keywords. 
I'm not sure if this was on propose or if something is missing.
Also could you please explain the 00024CppProxy?
Thanks! 👍

test-suite/generated-src/java/com/dropbox/djinni/test/Conflict.java
csimmons0: Java interfaces can't have private members. I read that they'll be able to in Java 9.

test-suite/generated-src/java/com/dropbox/djinni/test/ConstantsInterface.java
csimmons0: Fields in Java interfaces are implicitly public static final.

test-suite/generated-src/jni/NativeCppException.cpp
csimmons0: get() is a static function declared here in exception.djinni
```
cpp_exception = interface +c {
    throw_an_exception(): i32;
    static get(): cpp_exception;
}
```
This change moved the Java declarations of static functions that are implemented in C++ into the CppProxy inner class. "_00024" corresponds to the "$" symbol, which sits between inner and outer Java classes' names in the native name of a Java method that belongs to an inner class.

4brunu:
@csimmons0 Thanks for the quick responses, looks good to me 👍

test-suite/run_djinni.sh
```diff
@@ -95,6 +96,7 @@ fi
 "$base_dir/../src/run-assume-built" \
     --java-out "$temp_out_relative/java" \
     --java-package $java_package \
+    --java-generate-interfaces true \
     --java-nullable-annotation "javax.annotation.CheckForNull" \
     --java-nonnull-annotation "javax.annotation.Nonnull" \
     --java-use-final-for-record false \
```
artwyman: Need to enable Java interfaces on the 3rd Djinni run below (GitHub won't let me comment on a line which isn't changed), so that all generation is consistent.  I noticed in looking at diff since my last review that ExternalInterface1.java had reverted back to using `abstract class`.

artwyman:
Sorry for the delay again.  I think I'm fine with this feature as implemented now.  I only pointed out one problem with the test suite configuration in my latest review.

@xianwen will need to test this on the Dropbox codebase to ensure it doesn't break anything (at least when the feature is disabled).  It may also be worth considering enabling it too (and modifying existing code to use `implements` rather than `extends` when necessary.  I believe the Dropbox codebase has either already adopted Java 1.8, or will do so soon.

4brunu:
@artwyman @xianwen Any news on this PR? I'm really looking forward to see this get merged 🙂

artwyman:
See my last round of comments for the remaining small issues behind my "request changes".  I think it was only the test suite and state of re-generation thereof which was at issue.
In parallel @xianwen should be able to test this to make sure it doesn't break anything in the Dropbox codebase, since we use modes not covered in the test suite.

xianwen:
@4brunu, I was waiting for @csimmons0  to update the diff per: https://github.com/dropbox/djinni/pull/288/files/941947dc1345021988ac3af0072d9416b229b502..cd86c27df026193fe497918eadd064e39076d606#diff-2fa8e91b1f628c75dbbe9f7bebb317ee

I'll wait for one more week in case @csimmons0 prefers to take care of it, otherwise I'll adjust it myself next week, and merge it if it passed all our tests.

4brunu:
@artwyman @xianwen thanks for the update and the help to get this merged. 🙂

csimmons0:
Should be good now.

csimmons0:
@xianwen I believe that 7d0fc1d fixes the build error you hit.

gabrielformiga:
@xianwen any news on this pr? It's a very useful change on Java/Kotlin (Android) side, since it allows less verbose code. Thanks!

artwyman:
Looks like @xianwen may need to take a look at the latest commits here.
My recollection is that we decided it was okay for Java 8 to be required to enable this configuration flag, so long as with it disabled Djinni can still be used with older Java versions.

bilsou:
@xianwen Any update on this ? As @gabrielformiga mentioned, it would be really useful on on the Android/Kotlin side.

PR #289, opened 2016-11-10
Update ios version in Make targets to 10.1.

csimmons0:
From the Slack discussion channel: https://mobilecpp.slack.com/archives/general/p1478766914000338

In short, the Makefiles say that the iOS targets need to run on a simulator that's running ios 9.3, which causes the iOS targets to fail if you have the latest version of Xcode installed. This change updates the Makefiles to specify ios 10.1.


smarx:

@csimmons0, it looks like you've already signed the Dropbox CLA. Thanks!

PR #290, opened 2016-11-20
Add Windows Phone (C++/Cx) support

DEGoodmanWilson:
The code in this PR is over a year old. I don't rightly know how hard this is going to be to merge. I imagine it's not going to be easy—there'll be changes that need to be made. Let's talk! I'm ready to finally get this in.


smarx:

@DEGoodmanWilson, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@DEGoodmanWilson, thanks for signing the CLA!

gnichola:
@DEGoodmanWilson have you seen the blog MS released this morning?  I know it is early, but would this change your thinking or do you think Cx is still a valid way to go?
https://blogs.windows.com/buildingapps/2016/11/28/standard-c-windows-runtime-cwinrt

DEGoodmanWilson:
Well, that's pretty darned cool! I like Cx OK, but it's got its quirks (and was a huge step up on CLI), but this looks very very promising. That said, in the end, if you are developing with .NET assemblies, you might not care how those assemblies were built, and to the end user of djinni it might not make a whole lot of diffference, since _all_ of the Cx code is auto-generated, So…

DEGoodmanWilson:
(Also, I know I need to clean up the commit history a bit here! This was many months of effort, and there are some extraneous commits I need to remove from the PR)

AndrewAtAvenza:
I was actually about to start working on this but this looks pretty far along! I'm probably not the best person to fix the conflicts (I'm not even sure how since I definitely don't have write access) but I'd be happy to assist if possible. At the very least I could probably test this against our code base and see how that flies.

steipete:
We'd love to see this as well and I offer to help. We use djinni a lot and our C++ parts run on Windows as well - can try this branch with our setup as a larger test case once it's rebased - if that helps.

DEGoodmanWilson:
@steipete I've got this refactored so only the relevant contributions—and only _my_ contributions—are included. I can't see what the conflicts are, curiously. Git is not cooperating today. Anyway, would absolutely love your feedback.

support-lib/support_lib.gyp
```diff
@@ -43,5 +43,19 @@
                 ],
             },
         },
+        {
+            "target_name": "djinni_cx",
+            "type": "static_library",
+            "sources": [
+            ],
```
artwyman: Should list the header files, if nothing else.  Not sure if that's relevant to whichever generator you're using for this, but it's the normal standard used in the other targets in this file.

src/source/Main.scala
```diff
@@ -194,6 +214,37 @@ object Main {
         .text("Optional file in which to write the list of output files produced.")
       opt[Boolean]("skip-generation").valueName("<true/false>").foreach(x => skipGeneration = x)
         .text("Way of specifying if file generation should be skipped (default: false)")
+      opt[File]("cx-out").valueName("<out-folder>").foreach(x => cxOutFolder = Some(x))
+        .text("HAHAHAHAHAHAHAHAHAHA")
```
artwyman: Can you provide real documentation, please?

src/source/Main.scala
```diff
@@ -238,6 +298,10 @@ object Main {
       cppIdentStyle = cppIdentStyle.copy(enumType = cppTypeEnumIdentStyle)
     }
 
+//    if (cxTypeEnumIdentStyle != null) {
```
artwyman: Needed?

support-lib/cx/CppWrapperCache.h
```diff
@@ -0,0 +1,86 @@
+//
```
artwyman: Can the Cpp and Cx proxy caches be refactored to use the shared template impl in proxy_cache_*?

src/source/JNIGenerator.scala
```diff
@@ -31,9 +31,9 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
   val jniBaseLibClassIdentStyle = IdentStyle.prefix("H", IdentStyle.camelUpper)
   val jniBaseLibFileIdentStyle = jniBaseLibClassIdentStyle
 
-  val writeJniCppFile = writeCppFileGeneric(spec.jniOutFolder.get, spec.jniNamespace, spec.jniFileIdentStyle, spec.jniIncludePrefix) _
+  val writeJniCppFile = writeCppFileGeneric(spec.jniOutFolder.get, spec.jniNamespace, spec.jniFileIdentStyle, spec.jniIncludePrefix, spec.cppExt, spec.cppHeaderExt) _
```
artwyman: I don't see where these new args are passed.  Can you clarify what changes you're making to non-Cx generators, and how they're related?

phraemer:
Hi @DEGoodmanWilson ,

As @artwyman pointed out a working example would be great. I'm trying to figure out what changes are necessary to get the example working. Is it even complete enough for that?

I added something like...
```bash
cxcpp_out="$base_dir/generated-src/cxcpp"

...

   --cxcpp-out "$temp_out/cxcpp" \
   --cxcpp-namespace textsort_cx \
```
to `example/run_djinni.sh` but it seems like some headers are missing. For example that would generate `example/generated-src/cxcpp/SortItems_convert.h` and within the `textsort_cx::SortItems` class we get `using CxType = ::djinni::SortItems^;`
I assume that type in the `djinni` namespace would come from the missing `SortItems.h` header.
Perhaps I'm just missing a switch in the script?

phraemer:
Of course posting here what I tried prompted me to look closer at the new switches and I can see I needed `--cx-out "$temp_out/cxcpp" \` which gets me a step further. Don't know if that's enough yet.

Nevertheless if the example was fully working out of the box that would be very cool and I could get a better idea where I could perhaps contribute to this effort.

phraemer:
For anyone else trying this out here are a few hints to get the example generating code you can use in a Visual Studio project to build a WinRT component called `MyComponent` that bridges your native C++ with the managed languages, C# etc.

I added this to `run_djinni.sh`
```bash
cxcpp_out="$base_dir/generated-src/cxcpp"
cx_out="$base_dir/generated-src/cx"
...
    --cpp-namespace MyComponent_CPP \
...
    --cxcpp-out "$temp_out/cxcpp" \
    --cx-out "$temp_out/cx" \
    --cx-namespace MyComponent \
    --cxcpp-namespace MyComponent_CXCPP \
    --cpp-include-prefix "generated-src/cpp/" \
    --cx-include-prefix "generated-src/cx/" \
    --cxcpp-include-prefix "generated-src/cxcpp/" \
    --cxcpp-include-cpp-prefix "generated-src/cpp/" \
...
mirror "cxcpp" "$temp_out/cxcpp" "$cxcpp_out"
mirror "cx" "$temp_out/cx" "$cx_out"
```

You need different namespaces for `cx`, `cpp` and `cppcx` or you will have problems in your WinRT component. The `cx` namespace should be the same as the name of your component.

In `example.djinni`
```python
textbox_listener = interface +x {
```

In `handwritten-src/cpp/sort_items_impl.hpp`
```c++
#include "generated-src/cpp/sort_items.hpp"
#include "generated-src/cpp/textbox_listener.hpp"

namespace MyComponent_CPP {
```
In `handwritten-src/cpp/sort_items_impl.cpp`
```c++
namespace MyComponent_CPP {
```

PR #299, opened 2017-02-08
fix #298

### LINKED ISSUE #298: Windows: path not found error
When you start a djinni, under windows, if one of the command-line arguments contains the symbol < and >, we get the following error: "the system cannot find the path specified":

D:\7\service-communicator\djinni>sh d:\7\djinni\djinni\src\run --cpp-optional-he
ader "<boost/optional.hpp>" --idl  person/person.djinni
Building Djinni...
[info] Loading project definition from D:\7\djinni\djinni\src\project
[info] Set current project to src (in build file:/D:/7/djinni/djinni/src/)
[success] Total time: 0 s, completed 08.02.2017 17:30:09
[info] Wrote start script for mainClass := Some(djinni.Main) to D:\7\djinni\djin
ni\src\target\start.bat
[success] Total time: 0 s, completed 08.02.2017 17:30:09
Системе не удается найти указанный путь.

konovalov-aleks:


smarx:

artwyman:
Can you explain the shell magic here?  Is this removing escape characters added by the Windows cmd.exe?  As I mentioned in my comment on the issue here, I think it might not make sense to rely on proper arg passing from cmd to sh/bash, but instead write a bash script containing the args.  if this fix is assuming that MINGW implies "args are coming from cmd.exe" then that seems like it might be overbroad.

PR #300, opened 2017-02-26
Fixing multiple python issues

steweg:
Fixes the following:
 * unhandled exception during generation of python files
 * missing import in python generated files
 * CPyEnum.fromPy invalid assertion
 * allows python3 usage
 * import statements syntax (relative vs. system paths)


smarx:

artwyman:
A few requests before I go deep on the code review here:
- Can you provide some more context on the issues you found, and what was necessary to fix them?  For instance, I'm surprised you needed to do something for python3, since the example and test-suite both already run in Python3, and the generated code is intended to work for either language (and check at run-time when necessary).
- Can you re-generate the code for the example and test-suite?  This is both because I'd like to confirm that those run and pass for you, and also because the generated code is checked into the repo as an aid to code-review.  It's much easier to first review that the generated code looks reasonable, before reviewing the generator itself.

steweg:
Hi,

ok I have reviewed once again and also run my own tests regarding python3 option. It turned out that you are right, there was no need to adjust the code in that way. As this change is coming from colleague of mine, I will ask him for specific details why he asked me to adjust the code in that way, but for now I reverted that part.

For the rest, I have run "make example_python" and within Testing directory I run "make python" (I was not able to run full set of tests due to restrictions of my testing environments, but I guess those commands should be enough to trigger all needed steps you are looking for).

In order to explain the problems:
1. unhandled exception during generation of python files
During building of our djinni files, we have run into situation that several dh__ containers files were trying to be regenerated multiple times (with the one and the same content). This throws an exception, which was not handled and as such the generation failed. The root-cause here would be somehow limit the set of generated files to be uniq, so the exception would never be thrown, but I didn't find a successful way how to do it. Therefore I adjusted the code to handle such exception and simply skip regeneration as the file was already created

2. missing import in python generated files
in files like test-suite/generated-src/python/conflict_user.py, the proper imports directives are missing for e.g. line 40

3. CPyEnum.fromPy invalid assertion
The function is asserting "assert _ret.value != -1", which is not correct as "int" (type of "_ret") doesn't have "value" method. Python will complain about it when you reach the code during usage (apparently current tests doesn't catch it)

4. allows python3 usage ==> reverted

5. import statements syntax (relative vs. system paths)
This was done in order to allow one to properly create a package, that can be further on installed into the system (e.g. we are creating C++ library with python staff as one package). This is critical part as without the "." in front, the paths will not work properly. Unfortunately the "." in front causes python tests to fail due to:
"E   ImportError: attempted relative import with no known parent package"
This error is valid as python tests are not written in a way that they would be using "package". So question to you is whether you would like to have an option for "." to support packages or have it only with the form of without "."?

Anyway enjoy your vaction

artwyman:
A few clarifications:

1) Were you seeing the duplicate generation within a single run of Djinni, or multiple runs into the same generated output directory?  Would it be possible to provide an example project which demonstrates this issue?  I agree with you that it would be better to avoid duplicate generation, and maybe I can help to find a way to do that.

4) I'll need more time to dig deeper to have a preference on this issue, though my general approach in situations where different use cases and users have different needs is to add configuration via command-line options, so you might consider if that would make sense here, or if the test suite should simply be modified to better follow standard practices.

steweg:
Here is failing example:
```
MyRec = record {
        a: string;
}

IMyRecCallback = interface +p {
        MethodA(records: list<MyRec>);
        MethodB(records: list<MyRec>);
}
```

artwyman:
Some responses from your explanations and a look at your code:

1 (unhandled exception): Ah, is the problem simply that the logic to generate helpers for generic types (like `list<my_type>`) is triggered multiple times?  That should really be fixed in a more targeted way, by tracking which types have been generated, or by gathering all the types into a set before generation.  I think the try/catch approach is over-broad, and could hide a lot of other generation problems.

2 (missing import): I can see that in the example code now.  Seems straightforward.

3 (.value in assert): Seems straightforward too.  It would be good to see a test case which does exercise that code, as a way of validating the fix.

4 (reverted)

5 (import syntax): I don't know enough about various packaging/installer options for Python to know if there's a "right" answer here, so I'd prefer to see the code become more flexible, rather than switch to a different single answer based on your use case.  One way to make it generic might be to support an "import prefix" configurable on the Djinni command-line, which will be added before all imports of other Djinni classes.  That could be set to `""` for the current behavior, or `"."` for your desired behavior, or `"my.package.name"` to allow fully-qualified imports for people who may want that.  That also feels quite in keeping with existing Djinni behaviors which allow for prefixes on `#include` statements.

steweg:
Hi,

1A I reverted my original patch, created a test, add it into test suite and it turned out that it was actually a bug within checking inside "writtenFiles" hash map. This map included lowercase filename to actual filename, but when it was being checked, it was forgotten to lowercase first the key in order to find a match

1B After I find this issues, I solved it and adjust the test. It turned out that the testing of python was not failing properly due to missing exit code within Main (value other than 0) => so I added it. This part might need to be cherry-picked also into master as this is not related to python at all.

2-4 unchanged

5. I added string option for adding this kind of import prefix

I have rerun the tests and upload new patch-set

steweg:
@artwyman Hello, is there anything else that I need to do regarding this PR? Thanks

test-suite/djinni/foo_duplicate_file_creation.djinni
```diff
@@ -0,0 +1,13 @@
+Foo_Record = record {
```
artwyman: The names in this file should generally be `snake_case` to allow Djinni to do its standard style-selection for file-names and identifiers.  That's assumed when generating names, though not enforced.  I suspect that's why the bug you encountered was never seen before, since we hadn't tested your use case.

Given this test file is trying to verify the fix for this bug, I think it's right to keep its names mixed-case.  However I'd suggest adding a doc comment describing why this is the case, for future reference.  Otherwise I wouldn't be surprised if someone starts to enforce the name requirement, and comes across this test and gets confused by why it's here.

While you're at it, and explanation about the duplicate list types here, and what they're testing would be welcome.

artwyman:
Apologies for the lack of responsiveness.  I'm trying to get caught up on long-open PRs while ramping up a new maintainer.

test-suite/djinni/foo_duplicate_file_creation.djinni
steweg: Done

artwyman:
@xianwen ready to test and merge.

PR #304, opened 2017-03-17
Removing of unnecessary objc files generation

steweg:
This patch removes the unnecessary generation of ObjcppGenerator files related to interfaces, which are not meant to be implemented within OBJC nor CPP. These files were generated even wrongly as the generator expects that at least one side and generated code was not able to be compiled properly due to missing references etc...

Example of such interface:
```
MyInterface = interface +j {
  MyMethod1();
}
```

### DIFF
--- a/src/source/ObjcppGenerator.scala
+++ b/src/source/ObjcppGenerator.scala
@@ -62,6 +62,11 @@ class ObjcppGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
   def nnCheck(expr: String): String = spec.cppNnCheckExpression.fold(expr)(check => s"$check($expr)")
 
   override def generateInterface(origin: String, ident: Ident, doc: Doc, typeParams: Seq[TypeParam], i: Interface) {
+    if (!i.ext.objc && !i.ext.cpp) {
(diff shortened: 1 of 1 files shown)


smarx:

steweg:
After further testing it turned out that it is impossible to correctly compile after using my patch

steweg:
After further testing, this is impossible to be used like this. If I have CPP interface that expects some setting from Java, there is no way to tell this to djinni. Therefore the generated code of CPP will require as well "Private" Objc instance, therefore even if I have interface with just +j, I need also +o, otherwise I am not able to compile whole package (all generated Private Objc files)......... Most probable solution here would be to separate IDLs in a way that part of IDLs will be translated to Objc, Java and CPP, while another will be translated just for CPP and Java

artwyman:
Yes, I think that for the reasons you mention, cleanly avoiding unnecessary generation for single-language classes is tricky to do in a single generation run (if you want to to share C++ code).
Based that, I'm going to close out this PR, assuming you re-opened it by accident.  If I'm misinterpreting, re-open again and I can take closer look.

PR #308, opened 2017-04-06
Android parcelable support

4brunu:
Implemented support for android.os.parcelable

it's possible to apply this to the entire project via command-line `--java-implement-android-os-parcelable true` or per record `deriving(parcelable)`.

This PR is a continuation of https://github.com/dropbox/djinni/pull/257.
I picked the work done by @konovalov-aleks and apply the code review changes requested by @artwyman.

I don't claim the credits of this work to me, all the credit's should go to @konovalov-aleks.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -210,8 +210,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
 
     writeHppFile(cppName, origin, refs.hpp, refs.hppFwds, writeCppPrototype)
 
-    //If the only derivingType is AndroidParcelable, there is no need to generate the cpp file
-    if (r.consts.nonEmpty || (r.derivingTypes.nonEmpty && r.derivingTypes != Set(DerivingType.AndroidParcelable))) {
+    if (r.consts.nonEmpty || r.derivingTypes.contains(DerivingType.Eq) || r.derivingTypes.contains(DerivingType.Ord)) {
       writeCppFile(cppName, origin, refs.cpp, w => {
         generateCppConstants(w, r.consts, actualSelf)
 
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -210,7 +210,8 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
 
     writeHppFile(cppName, origin, refs.hpp, refs.hppFwds, writeCppPrototype)
 
-    if (r.consts.nonEmpty || r.derivingTypes.nonEmpty) {
+    //If the only derivingType is AndroidParcelable, there is no need to generate the cpp file
+    if (r.consts.nonEmpty || (r.derivingTypes.nonEmpty && r.derivingTypes != Set(DerivingType.AndroidParcelable))) {
       writeCppFile(cppName, origin, refs.cpp, w => {
         generateCppConstants(w, r.consts, actualSelf)
 
--- a/src/source/JavaGenerator.scala
+++ b/src/source/JavaGenerator.scala
@@ -473,10 +473,10 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
         case MOptional => {
           if (inOptional)
           	throw new AssertionError("nested optional?")
-          w.wl("if (in.readByte() == 0)").nested {
+          w.wl("if (in.readByte() == 0)").braced {
             w.wl(s"this.${idJava.field(f.ident)} = null;")
           }
-          w.wl("else").nested {
+          w.wl("else").braced {
             deserializeField(f, f.ty.resolved.args.head.base, true)
           }
         }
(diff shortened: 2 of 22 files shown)


smarx:

artwyman:
Thanks for resurrecting this.  No time to do a review yet (I'll get to it when I can), but I have a quick response to your description.  Is the cmd-line arg you suggested based on the needs of your own use case, or based on my suggestions in the prior code-review?  If I'm remembering correctly cmd-line arg I had in mind at the time would've enabled the feature for records marked deriving(parcelable), not for all records, since I think I was suggesting it as a way for a single `.djinni` file to support generation of Java-for-Android and Java-not-for-Android.

4brunu:
It was based in my needed's, because I want to enable this for all my records.
For Android projects I don't think there is disadvantages that all records implement parcelable, the only issue is if you want to support Android and a Java project simultaneously.
I don't mind to add `deriving(parcelable)` to the records since this is a robust and flexible approach, but the the command-line argument might be handy to Android projects only.
Do you think I should remove the command-line argument or keep it?

4brunu:
@artwyman I pushed some more fixes and optimizations and I think this is ready to code review.
The only thing I'm not sure is about the command-line argument, because equatable for example doesn't have a command-line argument to make all records equatable, so I'm not sure what's the right thing to do here.

4brunu:
@artwyman Sorry, I misread your comment. 
In the last commit the cmd-line arg enables the feature only for records marked deriving(parcelable).
This way it's possible to support Android and Java project's simultaneously with a single .djinni file.

4brunu:
@artwyman Do you have any news on this? :)

src/source/CppGenerator.scala
```diff
@@ -210,7 +210,8 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
 
     writeHppFile(cppName, origin, refs.hpp, refs.hppFwds, writeCppPrototype)
 
-    if (r.consts.nonEmpty || r.derivingTypes.nonEmpty) {
+    //If the only derivingType is AndroidParcelable, there is no need to generate the cpp file
+    if (r.consts.nonEmpty || (r.derivingTypes.nonEmpty && r.derivingTypes != Set(DerivingType.AndroidParcelable))) {
```
artwyman: It would be better if this explicitly checked for the derivings that it cares about, rather than only ruling out one of them.  If another new deriving is added in future, and it doesn't require editing this file to add new generated contents below, then it shouldn't require modifying this line either.

test-suite/djinni/primtypes.djinni
```diff
@@ -14,4 +14,4 @@ assorted_primitives = record {
     o_sixtyfour: optional<i64>;
     o_fthirtytwo: optional<f32>;
     o_fsixtyfour: optional<f64>;
-} deriving (eq)
+} deriving (eq,parcelable)
```
artwyman: nit: space after ,

test-suite/generated-src/java/com/dropbox/djinni/test/AssortedPrimitives.java
```diff
@@ -194,4 +194,111 @@ public String toString() {
...
+    public AssortedPrimitives(android.os.Parcel in)
+    {
+        this.mB = in.readByte() != 0;
+        this.mEight = in.readByte();
+        this.mSixteen = (short)in.readInt();
+        this.mThirtytwo = in.readInt();
+        this.mSixtyfour = in.readLong();
+        this.mFthirtytwo = in.readFloat();
+        this.mFsixtyfour = in.readDouble();
+        if (in.readByte() == 0)
```
artwyman: Please generate braces around ifs, even in generated code.  Missing them makes me nervous about unintended consequences when people insert lines later.

4brunu:
@artwyman I have made the changes that you asked. Please let me know if there are any more required changes

src/source/JavaGenerator.scala
```diff
@@ -534,11 +534,11 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
         case MOptional => {
           if (inOptional)
           	throw new AssertionError("nested optional?")
-          w.wl(s"if (this.${idJava.field(f.ident)} != null) {").nested {
+          w.wl(s"if (this.${idJava.field(f.ident)} != null)").braced {
```
artwyman: I think the old solution here was better because it generates `if (...) {` and `} else {` on single lines.  The current output is messily long.  Can you either revert to this behavior in both parts of this generator, or arrange to get the shorter output with braced?

4brunu:
@artwyman Done and I also improved the bracing for the whole parcelable feature to be more consistent with the rest of the generated code

test-suite/generated-src/java/com/dropbox/djinni/test/AssortedPrimitives.java
```diff
@@ -196,84 +196,66 @@ public String toString() {
...
         this.mFthirtytwo = in.readFloat();
         this.mFsixtyfour = in.readDouble();
-        if (in.readByte() == 0)
-        {
+        if (in.readByte() == 0) {
             this.mOB = null;
         }
-        else
-        {
+        else {
```
artwyman: Can you make this `} else {`?  I see that elsewhere.

4brunu:
@artwyman Done :)

4brunu:
@artwyman @xianwen Should this be documented on Readme?

artwyman:
Good idea.  Want to send a new PR?

4brunu:
I will, once I have some time 👍

PR #309, opened 2017-04-06
Cppfixes 201704

msjarrett:
Please consider the following changes - these fix build issues in our environment, and perhaps will help others.

- Adding :: to djinni namespace inside the DJINNI_ASSERT macros. This is already done in other macros in the file, and prevents a deeper namespace from being chosen inadvertently.
- proxy_cache_interface.hpp uses shared_ptr in a template class without including <memory>.
- Something about objectFromHandleAddress() is causing the compiler to segfault on the specified line. Not sure why (still investigating), but in the meantime, the temporary variable allows the compiler to complete.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,9 +437,6 @@ Run `make test` to invoke the test suite, found in the test-suite subdirectory. 
 - Iulia Tamas
 - Andrew Twyman
 
-## Contributors
-- Google Inc.
-
 ## Contacts
 - Andrew Twyman - `atwyman@dropbox.com`
 - Jacob Potter - `djinni@j4cbo.com`
--- a/README.md
+++ b/README.md
@@ -437,6 +437,9 @@ Run `make test` to invoke the test suite, found in the test-suite subdirectory. 
 - Iulia Tamas
 - Andrew Twyman
 
+## Contributors
+- Google Inc.
+
 ## Contacts
 - Andrew Twyman - `atwyman@dropbox.com`
 - Jacob Potter - `djinni@j4cbo.com`
--- a/support-lib/jni/djinni_support.hpp
+++ b/support-lib/jni/djinni_support.hpp
@@ -340,6 +340,8 @@ template <class T>
 static const std::shared_ptr<T> & objectFromHandleAddress(jlong handle) {
     assert(handle);
     assert(handle > 4096);
+    // Below line segfaults gcc-4.8. Using a temporary variable hides the bug.
+    //const auto & ret = reinterpret_cast<const CppProxyHandle<T> *>(handle)->get();
     const CppProxyHandle<T> *temp = reinterpret_cast<const CppProxyHandle<T> *>(handle);
     const auto & ret = temp->get();
     assert(ret);
--- a/support-lib/jni/djinni_support.hpp
+++ b/support-lib/jni/djinni_support.hpp
@@ -156,11 +156,11 @@ void jniThrowAssertionError(JNIEnv * env, const char * file, int line, const cha
 
 #define DJINNI_ASSERT_MSG(check, env, message) \
     do { \
-        djinni::jniExceptionCheck(env); \
+        ::djinni::jniExceptionCheck(env); \
         const bool check__res = bool(check); \
-        djinni::jniExceptionCheck(env); \
+        ::djinni::jniExceptionCheck(env); \
         if (!check__res) { \
-            djinni::jniThrowAssertionError(env, __FILE__, __LINE__, message); \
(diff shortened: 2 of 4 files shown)


smarx:

@msjarrett, it looks like you've already signed the Dropbox CLA. Thanks!

support-lib/jni/djinni_support.hpp
```diff
@@ -340,7 +340,10 @@ template <class T>
 static const std::shared_ptr<T> & objectFromHandleAddress(jlong handle) {
     assert(handle);
     assert(handle > 4096);
-    const auto & ret = reinterpret_cast<const CppProxyHandle<T> *>(handle)->get();
+    // Below line segfaults gcc-4.8. Using a temporary variable hides the bug.
+    //const auto & ret = reinterpret_cast<const CppProxyHandle<T> *>(handle)->get();
+    const CppProxyHandle<T> *temp = reinterpret_cast<const CppProxyHandle<T> *>(handle);
```
j4cbo: Name this "proxy_handle" or something more descriptive?

README.md
```diff
@@ -437,6 +437,9 @@ Run `make test` to invoke the test suite, found in the test-suite subdirectory.
 - Iulia Tamas
 - Andrew Twyman
 
+## Contributors
```
artwyman: @msjarrett can you explain this addition please?  I assume you're at Google, so I'd guess this isn't inaccurate, but it's far from complete since we haven't made a habit of calling out all individual people/companies which contribute in the past.

msjarrett: Our OSS policy asks us to add the company to the list of contributors, but as you mentioned, there isn't really a complete list in this project. The readme seems like the wrong place, but it'd be even weirder for me to create a new authors file for it.

Any suggestion for a better place for this?

artwyman: Hmm, yeah, maybe we should think about having an official list.  This is the first time someone's raised the issue of a policy like this.  Is the policy "you must add to the list" or can it be interpreted as "add to the list if it exists"?  It seems like in our current state our "contributors list" is `git log`, and I don't know if there are global norms on GitHub about doing otherwise.

artwyman: Is this coming from this policy: https://opensource.google.com/docs/patching/#how-to-patch

If so, that seems to suggest "If copyright authors are listed in a LICENSE, COPYING, AUTHORS", which wouldn't apply here.  We could work on changing that if you'd like, but may not need to delay this diff on it.

It may be that the current list of "Authors" is poorly labeled/named.  It's a list of "primary" contributors, not all contributors.  We've used it as a way to recognize people who wrote large chunks of the tool, or whole new features.  It's not intended to be a claim of copyright, since the whole project is copyrighted by Dropbox, per the CLA.

msjarrett: Yes, the above link is the basis of the policy. I can see how it could be interpreted either way, but my compliance team is insisting it applies for this particular change.
Is there somewhere I could put this that would be more suitable?

support-lib/jni/djinni_support.hpp
msjarrett: Done.

README.md
artwyman: Hmm, the consensus in the Dropbox open-source team is the opposite.  "If copyright authors are listed in a LICENSE, COPYING, AUTHORS or similar file" evaluates to false, so no action is required.  The only place our contributors are listed is in git history, which will be attributed when I merge this.  Can you possibly put your compliance team in touch with us?  Do you know if it's the Authors list which is at issue here, or do they think the addition is required even if it's the first one.

Anyhow, I'd rather not let legal wrangling actually get in the way of a good contribution.  I'd be open to possibly adding a CONTRIBUTORS file, with a paragraph at the top explaining it's not intended to be complete (the complete list being the git history), but only present where required for legal purposes.  I'd also be open to clarifying/renaming the Authors section in the README to make it clear it's a way of acknowledging "top contributors" not a list of all contributors.

msjarrett: Swapped the README.md change for a CONTRIBUTORS file. Does this work?

msjarrett:
Changed the README.md for a CONTRIBUTORS file.

msjarrett:
@artwyman Does the contributors change address your concern?

msjarrett:
Anybody....?

artwyman:
Hi @msjarrett.  Sorry for the long delay.  I didn't intentionally stonewall this.  It just came in at around the same time as I got too busy to keep up with Djinni for a while.  I don't love the CONTRIBUTORS file, but I'm willing to live with it to keep the lawyers happy.  Will merge shortly.

PR #310, opened 2017-04-12
[Android] Fix Djinni initialization crash in multi shared library case

Guillaume227:
Fix Djinni initialization crash in multi shared library case (i.e. djinni_support_lib is a shared library used by other shared libraries).

Crash is observed in Android Studio 2.3 with NDK+cmake setup when hitting assert statement: 
```
class JniClass {
public:
    static const C & get() {
        (void)s_initializer;
        assert(s_singleton); // Crashes here
        return *s_singleton;
    }
```

My understanding of the problem is that static_registration class template header is included in two different translation units in the shared libs and so results in bad initialization of the JniClassInitializer map.

Explicit template specialization of static_registration<void *, const JniClassInitializer>::get_map() in the cpp file is a way to force that get_map method to a single translation unit and so enforce its singleton nature.

With the exact same code, init crash was observed on Android 5 but not on Android 6 (tested in both case on a Nexus 5). I have no idea why it's a problem with one Android version and not the other.

### DIFF
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -46,9 +46,9 @@ JniClassInitializer::registration_vec JniClassInitializer::get_all() {
     return get_vec();
 }
 
-JniClassInitializer::JniClassInitializer(const std::function<void()> & init) {
+JniClassInitializer::JniClassInitializer(std::function<void()> init) {
     const std::lock_guard<std::mutex> lock(get_mutex());
-    get_vec().emplace_back(init);
+    get_vec().push_back(std::move(init));
 }
 
 void jniInit(JavaVM * jvm) {
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -45,14 +45,9 @@ JniClassInitializer::registration_vec JniClassInitializer::get_all() {
     return get_vec();
 }
 
-JniClassInitializer::JniClassInitializer(const std::function<void()> & init) : init(init) {
+JniClassInitializer::JniClassInitializer(const std::function<void()> & init) {
     const std::lock_guard<std::mutex> lock(get_mutex());
-    get_vec().emplace_back(this);
-}
-
-JniClassInitializer::~JniClassInitializer() {
-    const std::lock_guard<std::mutex> lock(get_mutex());
-    get_vec().erase(std::remove(get_vec().begin(), get_vec().end(), this), get_vec().end());
+    get_vec().emplace_back(init);
 }
 
 void jniInit(JavaVM * jvm) {
@@ -60,7 +55,7 @@ void jniInit(JavaVM * jvm) {
 
     try {
         for (const auto & initializer : JniClassInitializer::get_all()) {
-            initializer->init();
+            initializer();
         }
     } catch (const std::exception & e) {
         // Default exception handling only, since non-default might not be safe if init
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -28,12 +28,37 @@ namespace djinni {
 // Set only once from JNI_OnLoad before any other JNI calls, so no lock needed.
 static JavaVM * g_cachedJVM;
 
+JniClassInitializer::registration_set & JniClassInitializer::get_set() {
+    static JniClassInitializer::registration_set m;
+    return m;
+}
+
+std::mutex & JniClassInitializer::get_mutex() {
+    static std::mutex mtx;
+    return mtx;
+}
+
+JniClassInitializer::registration_set JniClassInitializer::get_all() {
+    const std::lock_guard<std::mutex> lock(get_mutex());
+    return get_set();
+}
+
+JniClassInitializer::JniClassInitializer(const std::function<void()> & init) : init(init){
+    const std::lock_guard<std::mutex> lock(get_mutex());
+    get_set().emplace(this);
+}
+
+JniClassInitializer::~JniClassInitializer() {
+    const std::lock_guard<std::mutex> lock(get_mutex());
+    get_set().erase(this);
+}
+
 void jniInit(JavaVM * jvm) {
     g_cachedJVM = jvm;
 
     try {
-        for (const auto & kv : JniClassInitializer::Registration::get_all()) {
-            kv.second->init();
+        for (const auto & initializer : JniClassInitializer::get_all()) {
+            initializer->init();
         }
     } catch (const std::exception & e) {
         // Default exception handling only, since non-default might not be safe if init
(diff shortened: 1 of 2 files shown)


smarx:

@Guillaume227, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@Guillaume227, thanks for signing the CLA!

4brunu:
I don't have enough knowledge to evaluate this PR, but I had the same issue, and this fixed it.
@artwyman Can you please take a look at this?
Thanks

support-lib/jni/djinni_support.cpp
```diff
@@ -28,6 +28,14 @@ namespace djinni {
 // Set only once from JNI_OnLoad before any other JNI calls, so no lock needed.
 static JavaVM * g_cachedJVM;
 
+template<>
```
artwyman: Can you add a comment explaining the rationale for this, similar to what's in this PR's description?

support-lib/jni/djinni_support.hpp
```diff
@@ -203,6 +203,10 @@ class JniClassInitializer {
     friend void jniInit(JavaVM *);
 };
 
+template<>
```
artwyman: Same desire for a comment here.

j4cbo:
So, static_registration's get_map() function is there to avoid running afoul of the ODR. We use static_registration in Djinni because it's a helper we had in another internal Dropbox library, and it made the most sense to just copy it into Djinni for the public release. But in Djinni, it's really only used for JniClassInitializer::Registration.

I think the ODR workaround (static variable in member function) in conjunction with shared libraries gives semantics that aren't quite right somehow, and that's what's causing the crashes people see.

It's the right call here to ditch the ODR hack and just move the definition of the unordered_map into djinni_support.cpp. But at that point, a general-purpose static_registration template is more confusing than it's worth. Since static_registration is only used one place in Djinni, I think it makes more sense to de-templatize it and turn it in to something like "JniClassRegistration", with a declaration in the header, and function definitions and the map and mutex defined in djinni_support.cpp.

Guillaume227:
Hi @j4cbo, thanks for explaining the history behind that static_registration design. I followed your advice and simplified it out. It occurred to me that JniClassInitializer could handle the static initialization on its own without an additionnal JniRegistration class. It that going too far ?

j4cbo:
That seems like a good way to do it. A few more requests:

- Use std::vector instead of unordered_set - we didn't really need the hash table behavior of unordered_map in the first place, just the ability to iterate over all elements, and std::vector is less code to pull in than unordered_set.

- The get_set/get_mutex functions are unnecessary now; you could just declare `static registration_vec m_vec;` inside JniClassInitializer in the header and define `/*static*/ JniClassInitializer::registration_vec JniClassInitializer::m_vec;` in djinni_support.cpp.

Also, please squash the changes down to one commit for merge.

Guillaume227:
Ok, set replaced with a vector as suggested.
I tried getting rid of the get/set but realized on my multi-lib test case that they are still needed, lest I get an assert in JniClass::get.

Regarding the squash isn't it something that should happen on your side when merging the PR, from what I gather there : https://github.com/blog/2141-squash-your-commits ?

artwyman:
Huh, I somehow missed it when GitHub started directly supporting squash rather than merge.  I'll try it out when this merges, so yes you shouldn't have to squash in advance.

support-lib/jni/djinni_support.cpp
```diff
@@ -28,12 +28,39 @@ namespace djinni {
...
+    static std::mutex mtx;
+    return mtx;
+}
+
+JniClassInitializer::registration_vec JniClassInitializer::get_all() {
+    const std::lock_guard<std::mutex> lock(get_mutex());
+    return get_vec();
+}
+
+JniClassInitializer::JniClassInitializer(const std::function<void()> & init) : init(init){
```
artwyman: format nit: space before {

support-lib/jni/djinni_support.cpp
```diff
@@ -28,12 +28,39 @@ namespace djinni {
...
+}
+
 void jniInit(JavaVM * jvm) {
     g_cachedJVM = jvm;
 
     try {
-        for (const auto & kv : JniClassInitializer::Registration::get_all()) {
-            kv.second->init();
+        for (const auto & initializer : JniClassInitializer::get_all()) {
+            initializer->init();
```
artwyman: This isn't different from the old version, but while we're simplifying, I wonder whether we really need the relationship between the static data and the individual JniClassInitializer objects.  Each one is just holding a std::function, and there's cleanup logic in ~JniClassInitializer but our use case never relies on that for static JniInitializer objects, and indeed there are some thread-safety concerns if ~JniClassInitializer ever raced with jniInit().

As an alternative, couldn't we just have a global vector<function<void()>>, and let the JniClassInitializer object insert into it, and then be a dead object after that point.  We could even register with a function not a class, but the syntax of the class is nice for declaration of code to run at static init time.  @j4cbo what do you think?

support-lib/jni/djinni_support.cpp
Guillaume227: space inserted, thanks.

support-lib/jni/djinni_support.cpp
j4cbo: Makes sense - storing a vector<function<void()>> instead of a vector<JniClassInitializer *> seems fine to me.

artwyman: @Guillaume227 can you go ahead and apply that simplification?  It would make me more comfortable, vs. the potential race-conditions we might now have with destructors, if used recklessly.

Guillaume227: Sure - I just pushed a change. 
Let me know if that covers what you had in mind.

support-lib/jni/djinni_support.cpp
```diff
@@ -28,12 +28,35 @@ namespace djinni {
...
+
+/*static*/
+JniClassInitializer::registration_vec JniClassInitializer::get_all() {
+    const std::lock_guard<std::mutex> lock(get_mutex());
+    return get_vec();
+}
+
+JniClassInitializer::JniClassInitializer(const std::function<void()> & init) {
+    const std::lock_guard<std::mutex> lock(get_mutex());
+    get_vec().emplace_back(init);
```
artwyman: One more thing noticed on final review.  This function should take its arg by value, and then use std::move here.  Since in the normal use case the input is a temporary std::function (created from a function pointer) this will avoid an extra copy (which involves heap allocation in the case of std::function).

Guillaume227: Good point ! Done.

artwyman:
Confirmed that this works for our app at Dropbox.  Merging shortly.

PR #313, opened 2017-05-04
Extern type defined with generic parameters incorrectly generated in C++

### LINKED ISSUE #149: Extern type defined with generic parameters incorrectly generated in C++
When defining an external type in yaml like so:

```yaml
name: generic_interface
typedef: 'interface'
params: [parameter]
prefix: ''
cpp:
  typename: 'test::GenericInterface'
  header: '"../generic_interface.hpp"'
  byValue: false
    
  # Followed by the java + obj-c definitions
```

And an interface using the type:
```
do_things(): generic_interface<another_type>;

```

The signature for a method using this type is generated thusly:

```c++
virtual std::shared_ptr<test::GenericInterface><AnotherType> doThings() = 0;
```

So the type parameter is tacked on at the end of the pointer definition, rather than after the type definition, as one would expect:

```c++
virtual std::shared_ptr<test::GenericInterface<AnotherType>> doThings() = 0;
```

rdegnan:
This commit fixes #149

### DIFF
--- a/src/source/CppMarshal.scala
+++ b/src/source/CppMarshal.scala
@@ -146,7 +146,7 @@ class CppMarshal(spec: Spec) extends Marshal(spec) {
       }
       withNs(ns, name)
     }
-    def base(m: Meta): String = m match {
+    def base(m: Meta, args: String): String = m match {
       case p: MPrimitive => p.cName
       case MString => if (spec.cppUseWideStrings) "std::wstring" else "std::string"
       case MDate => "std::chrono::system_clock::time_point"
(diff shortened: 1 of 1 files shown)


smarx:

@rdegnan, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@rdegnan, thanks for signing the CLA!

artwyman:
Hi @rdegnan.  Thanks for taking a stab at this.  Looks like this change breaks some existing functionality, though.  When I try to build the sample app I see many errors like this:

```
/Users/atwyman/src/djinni/example/generated-src/cpp/item_list.hpp:13:10: error: use of class template 'std::vector'
      requires template arguments
    std::vector items;

```

Looks like your change is affecting the handling of template args for types like vector<> in a bad way.

For future reference, when making any changes, please make sure you can `make all` from the top-level of the repo (which can only run on a Mac due to the inclusion of iOS stuff).  That'll re-generate and build the example apps and test suite, and run the tests.  Note that you should also check in the resulting generated code.  That's not a normal best-practice for generated code, but in this case, reviewing the generated code for the example and test suite is the easiest way for a reviewer like me to see the net effect of any changes to the generator.

PR #322, opened 2017-08-22
Fixed error when building docker image for unit test

choiip:
In CentOS 6 and Fedore 24, even the openjdk is officially updated, the softlink refer to the corresponding installed path.

In the debian jessie, problem solved by installing additional packages.


artwyman:
We don't use Djinni on Linux at Dropbox, and thus I don't have god context to evaluate this or test it out.  I wonder if @pwais (original author of the Docker support) would be willing to take a look.

pwais:
I could take a look this weekend; the Dockerfile patches look plausible to me.  (Thanks for the effort!)  If we had CircleCI for this repo we could definitely automate testing of the docker-based stuff.

PR #323, opened 2017-08-31
New "flags" archetype for enums oriented towards bitwise flags

choiip:
This change is originally committed by mknejp with resolved the conflict. I fixed issues pointed out by artwyman.


src/source/YamlGenerator.scala
```diff
@@ -103,11 +103,12 @@ class YamlGenerator(spec: Spec) extends Generator(spec) {
         }.mkString(" deriving(", ", ", ")")
       }
     }
-    td.body match {
-      case i: Interface => "interface" + ext(i.ext)
-      case r: Record => "record" + ext(r.ext) + deriving(r)
-      case e: Enum => "enum"
-    }
+  	td.body match {
```
artwyman: I'm still seeing tabs in the new version of this code, vs. the spaces in the old version.  Needs to be fixed to consistently use spaces (the preference for this repo).

src/source/generator.scala
```diff
@@ -386,6 +386,33 @@ abstract class Generator(spec: Spec)
     w.w(end)
   }
 
+  def normalEnumOptions(e: Enum) = e.options.filter(_.specialFlag == None)
+
+  def writeEnumOptionNone(w: IndentWriter, e: Enum, ident: IdentConverter) {
+    for (o <- e.options.find(_.specialFlag == Some(Enum.SpecialFlag.None))) {
```
artwyman: Naming the special flag NoneFlag or NoFlags might be clearer. I was tripped up for a bit by the use of None above, vs. the use of Enum.SpecialFlag.None here.

support-lib/jni/djinni_support.hpp
```diff
@@ -546,13 +546,45 @@ class JniEnum {
...
     JniEnum(const std::string & name);
+    jclass enumClass() const { return m_clazz.get(); }
 
 private:
     const GlobalRef<jclass> m_clazz;
     const jmethodID m_staticmethValues;
     const jmethodID m_methOrdinal;
 };
 
+class JniFlags : private JniEnum {
```
artwyman: Private inheritance is a reasonable compromise on the prior discussion here.  Should keep any external uses of this class from having any confusion.

test-suite/generated-src/cpp/access_flags.hpp
```diff
@@ -0,0 +1,56 @@
...
+    GROUP_EXECUTE = 1 << 5,
+    SYSTEM_READ = 1 << 6,
+    SYSTEM_WRITE = 1 << 7,
+    SYSTEM_EXECUTE = 1 << 8,
+    EVERYBODY = 0 | OWNER_READ | OWNER_WRITE | OWNER_EXECUTE | GROUP_READ | GROUP_WRITE | GROUP_EXECUTE | SYSTEM_READ | SYSTEM_WRITE | SYSTEM_EXECUTE,
+};
+constexpr access_flags operator|(access_flags lhs, access_flags rhs) noexcept {
+    return static_cast<access_flags>(static_cast<unsigned>(lhs) | static_cast<unsigned>(rhs));
+}
+inline access_flags& operator|=(access_flags& lhs, access_flags rhs) noexcept {
```
artwyman: Can you explain why some of these are inline while others are constexpr?  All were constexpr in the old version of this PR.  Is this a C++11 vs. C++14 limitation?

test-suite/java/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ file(
   ../generated-src/cpp/*.cpp
   ../handwritten-src/cpp/*.cpp)
 
-set(test_suite_common_flags "-g -Wall -Werror -std=c++11")
+set(test_suite_common_flags "-g -Wall -Werror -std=c++14")
```
artwyman: This should remain at c++11. Djinni requires C++11 as a minimum, and shouldn't require C++14.

test-suite/objc/DjinniObjcTest.xcodeproj/project.pbxproj
```diff
@@ -597,6 +619,7 @@
 				B5F06A7F1D4973BD005BE736 /* DBObjcOnlyListener.h */,
 				B5F06A801D4973BD005BE736 /* DBObjcOnlyListener+Private.h */,
 				B5F06A811D4973BD005BE736 /* DBObjcOnlyListener+Private.mm */,
+				CFBBB8EC1B6E3A67002EF52F /* DBAccessFlags.h */,
```
artwyman: Compared to the old PR I'm seeing some missing files in here.  Has this been tested?

test-suite/objc/DjinniObjcTest.xcodeproj/project.pbxproj
```diff
@@ -1105,6 +1146,7 @@
 		65868B6E1989FE4200D60EEE /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CLANG_CXX_LANGUAGE_STANDARD = "c++14";
```
artwyman: As above, this needs to stay at C++11.

test-suite/objc/DjinniObjcTest.xcodeproj/project.pbxproj
choiip: During I resolve the conflict, some of the files are removed in the master branch. 
I did the test by running "make objc" under the test-suite directory and got "** TEST SUCCEEDED **"

choiip: I found that I really miss 2 header files. I will fix it soon.

artwyman:
@xianwen can do the test & merge from here.

artwyman:
The CLA-bot doesn't seem to have triggered, but I manually confirmed that @choiip has signed the CLA.

xianwen:
Oops, `make all` passed on `master` branch, so I merged this commit.

But when I'm trying to merge this changes to `python` branch, I got the following errors:
```
Undefined symbols for architecture i386:
  "testsuite::FlagRoundtrip::roundtrip_empty_boxed(std::experimental::fundamentals_v1::optional<testsuite::empty_flags>)", referenced from:
      _cw__flag_roundtrip_roundtrip_empty_boxed in cw__flag_roundtrip.o
  "testsuite::FlagRoundtrip::roundtrip_access_boxed(std::experimental::fundamentals_v1::optional<testsuite::access_flags>)", referenced from:
      _cw__flag_roundtrip_roundtrip_access_boxed in cw__flag_roundtrip.o
ld: symbol(s) not found for architecture i386
```

Any ideas why this is happening and how to fix it?

artwyman:
That kind of thing can happen because Python isn't fully feature-complete on everything, and in particular nobody's implemented this flags feature for Python yet.  The general approach to that is the forking of the test suites, so in particular the `.djinni` input files for the test suite are separated into "common" files which all test suites use, and then some which are used only for non-Python, and others used only for Python.  My suspicion is that the new test cases for flags were included in the common set, and need to be moved into the non-Python set, just by moving things around in .djinni files.  If that doesn't work, and Python is actually triggering errors when parsing input which doesn't include any flags, then there might be a more complicated issue, but you should check that first.

PR #324, opened 2017-09-05
Add assembly to produce standalone jar.

### LINKED ISSUE #222: modern packaging / release for djinni executable jar
I've been using `run-assume-built` based upon the examples ( https://github.com/dropbox/djinni/blob/master/example/run_djinni.sh#L53 ) 

This can be a bit painful in a Docker context where I'd prefer to just run djinni using system-installed sbt and scala.  The `publish` target isn't super helpful right now... there's no main class defined in `build.sbt` for one thing.  I think `sbt` has come a long way since this project started.  What changes would we need to make a nice djinni executable?

yageek:
Using `sbt assembly` inside the `djinni` subdirectory allows to create a standalone `.jar`.

I am not a scala user. I tested the build scripts for the test-suite and the example and seems to work.

Could be a start point to fix #222.


.gitignore
```diff
@@ -34,3 +34,4 @@ intellij-plugin/djinni.jar
 
 # profiling output
 callgrind.out.*
+src/
```
artwyman: Can you explain this?  Ignoring the src/ directory doesn't seem obviously right.

yageek: You are completely right. I guess this is a wrong manipulation on my side during a find/replace sequence.

yageek:
I do not have a big experience as a Scala/SBT user. I have relocated the files to the [default structure](http://www.scala-sbt.org/0.13/docs/Howto-Customizing-Paths.html). I imagined the hierarchy would be more clean and the `assembly` would not require any configuration. I have modified the scripts `test-suite/run_djinni.sh` and `example/run_djinni.sh` accordingly and they seemed to work on my side.

Do you prefer I give another try with the previous files' organisation?

artwyman:
I'd prefer to at least treat file reorg as a separate change than adding a new assembly feature, which I'd expect to be pretty small and limited to the build files.  Reorging may not be a bad idea if it makes things more standard and clear, but that needs to be balanced against being a breaking change.

yageek:
I just updated the PR. This should switch back to the previous file hierarchy. The commans `sbt assembly` inside the `src` directory will produce a basic non versionned `.jar`.

yageek:
I updated the PR. Feel free to add comments.

artwyman:
@yageek ooks like our CLA-bot is broken, but it looks like we don't have a record of you signing the Dropbox CLA. Can you please sign it online here? https://opensource.dropbox.com/cla/

artwyman:
@xianwen ready to test and merge, once the CLA is signed.

artwyman:
CLA is signed, thanks @yageek

PR #326, opened 2017-09-21
Fix CFStringEncoding/NSStringEncoding confusion

### LINKED ISSUE #325: DJIMarshal+Private.h: CFStringEncoding vs. NSStringEncoding
I've hit a few compile errors in DJIMarshal+Private.h when trying to build Djinni-generated Obj-C interface code on Linux with GNUstep (yeah, yeah.) GNUStep provides `NSStringEncoding`, but _not_ `CFStringEncoding`, which is the return type of `getWCharEncoding` in DJIMarshal+Private.h, so the compiler takes issue with that.

However, looking at this code, I think it's slightly broken: `getWCharEncoding()` _declares_ a return value of `CFStringEncoding`, but the implementation _actually_ returns `NSStringEncoding` values. That return value is used in two places, both also in DJIMarshal+Private.h, on lines 157 and 165. The first calls `CFStringConvertEncodingToNSStringEncoding()` to "convert" the value (even though it's already actually an `NSStringEncoding`) and passes it to a message which expects an `NSStringEncoding`. The second (on line 165) just passes the return value directly to a message expecting an `NSStringEncoding`, casting it implicitly.

So, the use of `CFStringEncoding` here seems superfluous. It seems like `getWCharEncoding()` should simply return `NSStringEncoding`, and the call to `CFStringConvertEncodingToNSStringEncoding()` should be removed?

(Removing the references to `CFStringEncoding` also has the added bonus of unblocking building this code with GNUstep, as it happens...)

jgus:
Fixes #325

### DIFF
--- a/support-lib/objc/DJIMarshal+Private.h
+++ b/support-lib/objc/DJIMarshal+Private.h
@@ -119,14 +119,14 @@ struct String {
 };
 
 template<int wcharTypeSize>
-static CFStringEncoding getWCharEncoding()
+static NSStringEncoding getWCharEncoding()
 {
     static_assert(wcharTypeSize == 2 || wcharTypeSize == 4, "wchar_t must be represented by UTF-16 or UTF-32 encoding");
     return {}; // unreachable
 }
 
 template<>
-inline CFStringEncoding getWCharEncoding<2>() {
+inline NSStringEncoding getWCharEncoding<2>() {
     // case when wchar_t is represented by utf-16 encoding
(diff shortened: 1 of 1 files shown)


artwyman:
@jgus I forgot to check because the CLA-bot didn't trigger, but it looks like we don't have a record of you signing the CLA?  Can you please sign the CLA online?  You can find it here:

https://opensource.dropbox.com/cla/

If we don't get a signature, legal is going to make me revert this.

jgus:
@artwyman Done!

artwyman:
Confirmed, thanks!

PR #328, opened 2017-09-26
Python tooling issues

### LINKED ISSUE #327: Python compiler warnings from -Wstrict-prototypes
This is a new warning introduced in clang with Xcode 9 and enabled.  It's pointing out that in C (unlike C++) `foo()` is a function with unspecified arguments, not a function with no arguments.  `foo(void)` is the correct form in C (and works in C++ too).

It looks like this warning affects both support-library code and generated code, so will need some changes to the generator.

```
In file included from PyCFFIlib_cffi.c:432:
./clean_headers.h:42:82: warning: this function declaration is not a prototype [-Wstrict-prototypes]
struct DjinniPythonExceptionHandle * djinni_from_python_check_and_clear_exception(); // called in python
                                                                                 ^
                                                                                  void
./clean_headers.h:155:81: warning: this function declaration is not a prototype [-Wstrict-prototypes]
void list_string_add_callback__python_create(struct DjinniObjectHandle *( * ptr)());
                                                                                ^
                                                                                 void
2 warnings generated.
```

artwyman:
Fixing two issues here, both related to building Python on a Mac.  The second issue had to be fixed to let me test the fix for the first.

1. Xcode 9 has a new warning for C in which (unlike C++) `foo()` (meaning unspecified args) vs. `foo(void)` (meaning no args).  I altered the support-lib and generator to always use `(void)`, which works fine in C++ too.

2. The dylib we use in our example app won't load because new security features have disabled DYLD_LIBRARY_PATH.  I hacked around it by using an explicit library load path.

Fixes #327

### DIFF
--- a/example/run_py_example.sh
+++ b/example/run_py_example.sh
@@ -1,5 +1,5 @@
 #! /usr/bin/env bash
-set -e
+set -eux
 
 # Locate the script file.  Cross symlinks if necessary.
 loc="$0"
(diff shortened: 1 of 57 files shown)


artwyman:
@xianwen to review.  (I can't assign ownership yet since he hasn't accepted his group invite yet.)

example/run_py_example.sh
```diff
@@ -16,8 +16,16 @@ base_dir=$(cd "`dirname "$loc"`" && pwd)
 
 python_cmd=$1
 
+export PYTHONPATH=${PYTHONPATH:-""}
 export PYTHONPATH="$base_dir/../support-lib/py:$base_dir/../build_py/cffi:$base_dir/generated-src/python:$PYTHONPATH"
-export DYLD_LIBRARY_PATH=$base_dir/../build_py/cffi:$DYLD_LIBRARY_PATH
-(cd "$base_dir/../build_py/cffi" \
+# The use of install_name_tool below is a hack to force MacOS to know where to load libtextsort_py.dylib,
+# needed because DYLD_LIBRARY_PATH is now ignored by System Integrity Protection.  A real installed which
```
xianwen: > A real installed which

"A real installed **application** which" ?

example/run_py_example.sh
```diff
@@ -16,8 +16,16 @@ base_dir=$(cd "`dirname "$loc"`" && pwd)
...
+# used a dylib like this would need a cleaner solution based on @rpath or somesuch thing.  The file existence
+# checks before the commands look for the two different .so names produced for CPython 2 or 3.
+#export DYLD_LIBRARY_PATH=$base_dir/../build_py/cffi:$DYLD_LIBRARY_PATH
+cffi_dir=$(cd "$base_dir/../build_py/cffi" && pwd)
+(cd "$cffi_dir" \
         && "$python_cmd" ../../example/generated-src/cffi/pycffi_lib_build.py ../../support-lib/cwrapper/wrapper_marshal.h $(ls ../../example/generated-src/cwrapper/*.h) \
+        && echo && ([ ! -f PyCFFIlib_cffi.so ] || install_name_tool -change /usr/local/lib/libtextsort_py.dylib "$cffi_dir/libtextsort_py.dylib" PyCFFIlib_cffi.so) \
```
xianwen: I'm not quite familiar with `install_name_tool`.
Looks like if the `install_name_tool` has been executed, the test for `-f PyCFFIlib_cffi.so` would then fail?

I'm wondering if there is a better alternative for this testing, the `-f` test seems not quite robust.

PR #329, opened 2017-09-26
Upgrade the example to Gradle 4.2 and Android plugin 2.3.3

sschuberth:
Along with the necessary build system adjustments.


sschuberth:
Also, do not check for the NDK during Gradle configurion phase, otherwise a simple "gradle tasks" would fail if no NDK is configured. Do the checks later when the ndkBuild task is actually executed.

artwyman:
Leaving this to @xianwen to review and merge, but my high-level take is that since this is for the sample app, the risk of any compatibility-breakage is minimal.  So long as this works, and depends on GA versions of tools all Android devs are likely to have (i.e. not beta/preview versions) it should be fine.

artwyman:
@sschuberth looks like our CLA-bot is broken, but it looks like we don't have a record of you signing the Dropbox CLA.  Can you please sign it online here?  https://opensource.dropbox.com/cla/

sschuberth:
@artwyman That link seems to be dead. (Which might be the reason your CLA bot does not work?)

sschuberth:
Still dead for me:

```
Server not found

Firefox can’t find the server at opensource.dropbox.com.
```

sschuberth:
Nevermind, seems to be an intranet issue.

artwyman:
Sadly some companies block Dropbox for security reasons (misguided ones if you ask me :).
Did you manage to get through?  I don't see a signature in the back-end yet.

sschuberth:
I first need to the get CLA reviewed, which will take some more time, I'm sorry.

sschuberth:
@artwyman I've signed the CLA now!

artwyman:
Confirmed the CLA, thanks!  @xianwen can take it from here.

sschuberth:
Any update @xianwen?

xianwen:
Sorry for the delay, I'll take a look at it today, and get back to you once I finished the review.
Thanks for your patience.

example/android/gradlew
```diff
@@ -1,25 +1,43 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
xianwen: Could you explain why we should prefer `sh` over `bash`?
I'm used to go for `bash` as default.

sschuberth: Please note that this file is automatically created by the Gradle `wrapper` task. As to why the Gradle devs decided to go for `sh`, I guess that's for better portability across platforms.

example/android/gradlew.bat
```diff
@@ -60,11 +59,6 @@ set _SKIP=2
 if "x%~1" == "x" goto execute
 
 set CMD_LINE_ARGS=%*
-goto execute
-
-:4NT_args
-@rem Get arguments from the 4NT Shell from JP Software
-set CMD_LINE_ARGS=%$
```
xianwen: Why the support for `Take Command Console` removed here?
AFAIK, it's not deprecated. As I saw its latest stable releases is 20.11 on March 10, 2017, 6 months ago. And it's supporting Windows 10, the latest Windows release.

sschuberth: Please see my comment above, this file is also auto-generated by the `wrapper` task and thus comes straight from Gradle.

example/android/gradlew
```diff
@@ -40,31 +59,11 @@ case "`uname`" in
...
+  NONSTOP* )
+    nonstop=true
+    ;;
 esac
 
-# For Cygwin, ensure paths are in UNIX format before anything is touched.
-if $cygwin ; then
-    [ -n "$JAVA_HOME" ] && JAVA_HOME=`cygpath --unix "$JAVA_HOME"`
-fi
-
```
xianwen: Why line 45 ~ 49 removed here? It seems it's needed in order to support `cygwin`.

example/android/gradlew.bat
xianwen: Oops, I just removed my comments since I'v figured out that these 2 files are auto-generated.

xianwen:
LGTM, `make all` passed after testing it on latest master branch

PR #339, opened 2017-11-27
Upgrade example build

sschuberth:
Please have a look at the individual commit messages for the details.

This is a follow-up to PR #329.


artwyman:
Can you explain where/why the null ndkDir would occur?  I don't fully understand what the outcome would be if it's null, and thus the NDK build ends up being disabled.  If that happened, it would invalidate the usefulness of a passing build as a regression test.

I'll leave @xianwen to do the final test & merge.

FYI, we've switched to mostly doing merges via Squash, so a PR summary is useful for forming the final commit message.

Also FYI since the CLA-bot is broken, I've confirmed that @sschuberth has signed the CLA.

sschuberth:
> Can you explain where/why the null ndkDir would occur?

Simply if the NDK is not installed (but only the SDK). Granted, the example won't build in this case, but one might still want to be able to inspect the project e.g. by running `./gradlew tasks`. Currently, this would fail because the NDK directory is being relied on to be not `null` already when just configuring (not executing) the tasks.

sschuberth:
Ping @xianwen!

example/android/app/build.gradle
```diff
@@ -27,16 +26,22 @@ android {
...
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation 'com.google.code.findbugs:jsr305:3.0.0'
 }
 
 task ndkBuild(type: Exec) {
-    File ndkDir = project.android.ndkDirectory
-    def ndkBuildExecutable = new File(ndkDir, 'ndk-build')
+    def ndkDir = project.android.ndkDirectory
+    if (ndkDir) {
+        def ndkBuildExecutable = new File(ndkDir, 'ndk-build')
```
xianwen: This "ndkBuildExecutable" here is also used by "line 50" below, and wrapping it in the `if` statement makes it invisible outside and cause error when executing "lien 50": `if (!ndkBuildExecutable.exists()) {`

xianwen:
My apologies for the delayed review, there were too many things going on for me during last month.  Hopefully I'll be able to respond sooner from now on.

sschuberth:
Thanks for the review @xianwen and for pointing out that valid issue. I've fixed that plus made some updates to the other commits.

PR #341, opened 2017-12-05
Add a script to build static fat binary lib.

yageek:
Based on the platform file provided by the [ios-cmake](https://github.com/leetal/ios-cmake) repository, I added a script helping to generate a fat static library for apple platforms.
This only builds the OBJC part of the support-lib for now.

It basically loops over an array of valid platforms and uses `lipo` to merge all the different generated `.a` files. 

Those files are generated inside the `out` directory. 
The generated universal library is located at `out/libdjinni_support_lib_UNIVERSAL.a`


artwyman:
Logistics first: @yageek I don't think you've contributed before, so I need you to sign the CLA: https://opensource.dropbox.com/cla/

Our CLA-checking bot is broken, but just post here when you've signed and I can confirm your username the back-end.  Or if you're with an organization which signed let me know which to look for.

README.md
```diff
@@ -492,6 +492,21 @@ java -jar $DJINNI_JAR_DIR/djinni-assembly-0.1-SNAPSHOT.jar \
 
 *Note*: The `all` target of the main `Makefile` includes the `djinni_jar` target.
 
+## Generate an apple universal binary
+
+The `mac-build-support-lib.sh` helps you to build an universal static library.
+It uses the platform file of the [ios-cmake](https://github.com/leetal/ios-cmake) repository.
+
+It basically creates one universal static library per `IOS_PLATFORM` variable and uses `lipo`
```
artwyman: The combination of "Mac" in the script name and "IOS" here makes me wonder how general this feature is.  Does it cover all Apple platforms, or just iOS?

mac-build-support-lib.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/sh
+
+# This script helps you to build apple fat binaries.
+# This uses the excellent https://github.com/leetal/ios-cmake cmake platform file.
```
artwyman: Can you document somewhere exactly which version of that file was incorporated (by version if that exists, or else a git commit).

README.md
yageek: The name is bad, I have to admin. The generated binaries are for iOS platform, but to be compiled from a mac. I will rename the file.

mac-build-support-lib.sh
yageek: I'll add the sha-1 of the file inside the script.

yageek:
I did not have to time to dedicate to create a full CMake script/module to support fat binary. As some other platform file were existing, I took the quickest path to build the universal library.

The process is explained here: https://github.com/ruslo/sugar/wiki/Building-universal-ios-library

I never used CMake to create some kinde of loop over the same target and rebuild for different architecture. I guess a full CMake strategy is possible but I never tried. 

Do you still want to consider adding the script or do you prefer to concentrate effort on a full CMake version?

artwyman:
I wouldn't mind adding an optional helper script, so long as it's clear in docs what it does and doesn't do.  It might be best kept in community-supported state, meaning don't add it to `make all` but just let it be used by people who need/want it.  The normal approach to the support-lib in Xcode is just to add it as a source along with the generated sources, but a separate binary may certainly be useful to some people.

ios-build-support-lib.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/sh
```
artwyman: Can this file go somewhere other than at the top-level?  I feel like it belongs in `support-lib/`.

yageek: I have created the script at the same level as the `CMakeLists.txt` that builds the support library. Should I move it also?

artwyman: Hmm, good point, which I didn't realize since I don't use CMake.  If I were doing it from scratch, I feel like that CMakeLists.txt file belongs in the subdir too, along with the .gyp file.  I'd still suggest moving the script, though.  If we had some sort of a `bin` or `scripts` dir it would belong there, but `support-lib` seems best for now.

artwyman:
I can confirm now that you've signed the CLA, so you're all set.  Sorry for the confusion.

yageek:
No worry :) I was just wondering. I updated the script and moved it to the library subdir

prsolucoes:
Only to comment, i made a tool that have this feature: www.ezored.com

PR #342, opened 2017-12-07
Generate an XCode like bridging header.

yageek:
Generates also the version number and name variable inside the bridging header.


artwyman:
I think Xianwen should be the primary reviewer here since he has more experience with Swift.  Two bits of housekeeping though:

1) Can you please re-generate the code for the example and test-suite so they're in-sync, and reviewers can see the effects of your change?
2) Before this can merge you need to sign the CLA here: https://opensource.dropbox.com/cla/

yageek:
I think I have already signed it two times. Should we sign it for every PR?

artwyman:
Oops, sorry.  I think it was my mistake searching for your userid.  It's manual and error-prone since our bot broke.
I can confirm now that you've signed the CLA, so you're all set.

example/generated-src/objc/TextSort-Bridging-Header.h
```diff
@@ -1,6 +1,17 @@
...
 
+// TextSort_Bridging_Header.h
+// TextSort_Bridging_Header
+
+#import <Foundation/Foundation.h>
+
+//! Project version number for TextSort_Bridging_Header.
+FOUNDATION_EXPORT double TextSort_Bridging_HeaderVersionNumber;
+
+//! Project version string forTextSort_Bridging_Header.
```
xianwen: Missing space between "for" and "TextSort_Bridging_Header" here.

src/source/SwiftBridgingHeaderGenerator.scala
```diff
@@ -38,8 +38,23 @@ class SwiftBridgingHeaderGenerator(spec: Spec) extends Generator(spec) {
...
+    writer.write("// " + varPrefix + ".h\n")
+    writer.write("// " + varPrefix + "\n\n")
+  }
+
+  def writeBridgingVars(name: String, writer: Writer) {
+    val varPrefix = bridgingVarPrefix(name)
+    writer.write("#import <Foundation/Foundation.h>\n\n")
+    writer.write("//! Project version number for " + varPrefix +".\n")
+    writer.write("FOUNDATION_EXPORT double " + varPrefix + "VersionNumber;\n\n")
+    writer.write("//! Project version string for" + varPrefix +".\n")
```
xianwen: Need to add a space after `for`.

example/generated-src/objc/TextSort-Bridging-Header.h
```diff
@@ -1,6 +1,17 @@
 // AUTOGENERATED FILE - DO NOT MODIFY!
 // This file generated by Djinni
 
+// TextSort_Bridging_Header.h
+// TextSort_Bridging_Header
+
+#import <Foundation/Foundation.h>
+
+//! Project version number for TextSort_Bridging_Header.
+FOUNDATION_EXPORT double TextSort_Bridging_HeaderVersionNumber;
```
xianwen: I'd prefer to have a `_` separating `TextSort_Bridging_Header` and `VersionNumber` so that it's `TextSort_Bridging_Header_VersionNumber`.

src/source/SwiftBridgingHeaderGenerator.scala
xianwen: I don't like the way this variable is named. In this case: `"..." + varPrefix +"..."`, it's not used as a prefix.

Although it happens to be a prefix in the case of `VersionNumber`, I'd still prefer to remove the `prefix` in the var name, to make it something like `bridgingHeaderVarName`.

src/source/SwiftBridgingHeaderGenerator.scala
```diff
@@ -38,8 +38,23 @@ class SwiftBridgingHeaderGenerator(spec: Spec) extends Generator(spec) {
 }
 
 object SwiftBridgingHeaderGenerator {
-  def writeAutogenerationWarning(writer: Writer) {
+  
+  val bridgingVarPrefix = (s: String) => s.split('-').mkString("_")
```
xianwen: Similar to my comment below, I'd prefer to remove the `Prefix` here, to make it something like `bridgingHeaderName`, since it doesn't have to be used as a prefix. (Although it's not a great name either, as `-` is replaced by `_` and this method name I suggest wouldn't make it obvious.)

example/generated-src/objc/TextSort-Bridging-Header.h
yageek: I choosed the Xcode Convention. By calling a framework `MyLib`, the two variables are named `MyLibVersionNumber` and `MyLibVersionString`

src/source/Main.scala
```diff
@@ -346,6 +346,7 @@ object Main {
       objcppNamespace,
       objcBaseLibIncludePrefix,
       objcSwiftBridgingHeaderWriter,
+      objcSwiftBridgingHeader,
```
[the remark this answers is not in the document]
yageek: The variable is declared as `objcSwiftBridgingHeader` inside `main.scala`.
Do you want me to change the name of the property in `generator.scala`  from `objcSwiftBridgingHeaderName` to `objcSwiftBridgingHeader` instead?

example/generated-src/objc/TextSort-Bridging-Header.h
xianwen: I see, thanks for the explanation.
In that case, would it make sense to name it `TextSortBridgingHeaderVersionNumber`?

yageek: I guess yes 🤔

src/source/Main.scala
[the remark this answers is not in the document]
xianwen: Oh, I see, it's just the variable name here while `objcSwiftBridgingHeaderName` is the parameter, sorry I got confused.
Yes, I'd prefer to have the same name since this is the case for all the other parameters.

xianwen:
Thanks for your updates, I'll merge it after testing.

Issue #334, opened 2017-10-04
Merge #323 into python branch


xianwen:
#323 has been merged into master, but it can not be merged into python branch, because of the following error:

```
Undefined symbols for architecture i386:
  "testsuite::FlagRoundtrip::roundtrip_empty_boxed(std::experimental::fundamentals_v1::optional<testsuite::empty_flags>)", referenced from:
      _cw__flag_roundtrip_roundtrip_empty_boxed in cw__flag_roundtrip.o
  "testsuite::FlagRoundtrip::roundtrip_access_boxed(std::experimental::fundamentals_v1::optional<testsuite::access_flags>)", referenced from:
      _cw__flag_roundtrip_roundtrip_access_boxed in cw__flag_roundtrip.o
ld: symbol(s) not found for architecture i386
```

Needs to fix it so that it can be merged into python branch

PR #346, opened 2018-01-04
Separate `flags` to `enum_flags.djinni`

xianwen:
Separate `flags` to `enum_flags.djinni`, to exclude it from
`common.djinni`, so that it wouldn’t be picked up by Python  in
`python` branch.

Otherwise it causes failure in `python` branch since flags feature is
not implemented for Python.


xianwen:
This PR addresses #334

PR #347, opened 2018-01-05
Fixes error: expression is not a constant expression

freitass:
In GCC 7.2, the bitwise assignment operators generated for C++ flags can not be constexpr, causing the build to fail. This patch makes them inline instead. Also, the second argument is made const to silence some diagnostics in Visual Studio.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -80,7 +80,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
           w.w(s"constexpr $self operator$op($self lhs, $self rhs) noexcept").braced {
             w.wl(s"return static_cast<$self>(static_cast<$flagsType>(lhs) $op static_cast<$flagsType>(rhs));")
           }
-          w.w(s"inline $self& operator$op=($self& lhs, const $self rhs) noexcept").braced {
+          w.w(s"inline $self& operator$op=($self& lhs, $self rhs) noexcept").braced {
             w.wl(s"return lhs = lhs $op rhs;") // Ugly, yes, but complies with C++11 restricted constexpr
           }
         }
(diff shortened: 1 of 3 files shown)


artwyman:
Before merging this I'd like to hear commentary from @choiip who originally created this code in PR #323.

prsolucoes:
Hi, can you merge? I have the same problem:
https://travis-ci.org/ezored/dependency-logger/jobs/437848402

artwyman:
We can't legally merge this PR because @freitass hasn't signed the CLA.  We prefer to give authors the chance to take credit for their fixes, but given this one looks to be very simple/trivial code changes, and hasn't been followed up for many months, it might be reasonable for some other contributor to make an equivalent fix in a separate PR.

freitass:
Hi @xianwen, would you consider merging this PR again? I just signed the CLA, apologies for the delay.

src/source/CppGenerator.scala
```diff
@@ -80,7 +80,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
           w.w(s"constexpr $self operator$op($self lhs, $self rhs) noexcept").braced {
             w.wl(s"return static_cast<$self>(static_cast<$flagsType>(lhs) $op static_cast<$flagsType>(rhs));")
           }
-          w.w(s"constexpr $self& operator$op=($self& lhs, $self rhs) noexcept").braced {
+          w.w(s"inline $self& operator$op=($self& lhs, const $self rhs) noexcept").braced {
```
artwyman: I didn't previously notice the `const` here, and it doesn't appear to show up in the generated code below.  Was this added late?

If I'm understanding this correctly, the `const` here seems unnecessary since flags are passed by value anyway, so it's meaningless to the caller.  I prefer to avoid passing/returning by "const value" since it clutters the interface unnecessarily.

artwyman:
Thanks for the update.  Approved, and leaving this for @xianwen to merge in his next round of updates.

PR #348, opened 2018-01-11
Silence some optional Xcode 9 warnings

steipete:
This fixes an Analyzer warning that shows up in Xcode 9:
https://github.com/llvm-mirror/clang/blob/master/test/Analysis/number-object-conversion.cpp

```
34:20: warning: Converting a pointer value of type 'ObjcType' to a primitive boolean value; instead, either compare the pointer to nil or call -boolValue
            assert(x);
```

Also fixes an optional warning:

The `__unused` on the exception is needed when `-Wunused-exception-parameter` is enabled.

```
warning: unused exception parameter 'e' [-Wunused-exception-parameter]
        DJINNI_TRANSLATE_EXCEPTIONS()
        ^
```

Both are trivial changes that do not change functionality.
Thanks for making Djinni! We from PSPDFKit love it!

### DIFF
--- a/support-lib/objc/DJIError.h
+++ b/support-lib/objc/DJIError.h
@@ -30,6 +30,6 @@ namespace djinni {
     ::djinni::throwUnimplemented(__PRETTY_FUNCTION__, msg);
 
 #define DJINNI_TRANSLATE_EXCEPTIONS() \
-    catch (const std::exception & e) { \
+    catch (__unused const std::exception & e) { \
         ::djinni::throwNSExceptionFromCurrent(__PRETTY_FUNCTION__); \
     }
--- a/support-lib/objc/DJIMarshal+Private.h
+++ b/support-lib/objc/DJIMarshal+Private.h
@@ -29,7 +29,7 @@ struct Bool {
 
     struct Boxed {
         using ObjcType = NSNumber*;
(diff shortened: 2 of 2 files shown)


support-lib/objc/DJIError.h
```diff
@@ -30,6 +30,6 @@ namespace djinni {
     ::djinni::throwUnimplemented(__PRETTY_FUNCTION__, msg);
 
 #define DJINNI_TRANSLATE_EXCEPTIONS() \
-    catch (const std::exception & e) { \
+    catch (__unused const std::exception & e) { \
```
artwyman: Wouldn't it be better to simply delete the name?  It is truly unused, and removing the name is a more standard way to do it in C++.

support-lib/objc/DJIMarshal+Private.h
```diff
@@ -29,7 +29,7 @@ struct Bool {
 
     struct Boxed {
         using ObjcType = NSNumber*;
-        static CppType toCpp(ObjcType x) noexcept { assert(x); return Bool::toCpp([x boolValue]); }
+        static CppType toCpp(ObjcType x) noexcept { assert(x != nil); return Bool::toCpp([x boolValue]); }
```
artwyman: Interesting.  Stylistically I generally do prefer to let pointers be treated as bools directly, but in this context I'm fine with changing the code to be more generally usable by people who don't like that style and turn on warnings against it.

Issue #332, opened 2017-10-03
Keep parameter names in interface method docstrings in sync


louiswins:
It's always nice to have documentation, except when it's incorrect or out of date. Because the format of interface method parameters can differ between languages (by default `some_parameter` in C++ and `someParameter` in Objective-C and Java), your documentation will often be automatically incorrect in at least one language.

It would be nice to have Djinni update interface method docstrings when it updates the parameter names. At minimum, I would love it to rewrite Javadoc/Doxygen-style `@param parameter_name` in the comments to `@param parameterName` as appropriate, so that documentation in XCode and IntelliJ is correct and I stop getting warnings. It seems to me like it would also be safe to rewrite all references in the method docstring (because how often will you _want_ to refer to `some_param` when your method actually takes `someParam`) but that does incur more risk of false positives.

louiswins:
This was fixed for my purposes by #352, so I'm closing the issue.

PR #352, opened 2018-01-30
Follow naming convention for parameter names in method docstrings

louiswins:
The problem here is that if I have differing naming conventions in different languages then if I mention a method parameter name in my djinni docstring it will be wrong in at least one language (e.g. `some_parameter` in C++ but `someParameter` in Obj-C and Java). And I will probably mention them if I'm using doxygen or javadoc (`@param some_parameter`). Beyond being annoying, this causes warnings in XCode and probably other environments.

The fix will run every method docstring through a series of regexes that replace each djinni parameter name with the corresponding name in the language naming convention.

### Drawbacks
- I don't try to parse doxygen semantics or anything, I just blindly replace. But I don't expect that there are many cases where a parameter name with an underscore is used in a method docstring to mean something that is not actually the parameter name.
- I don't replace non-parameter names, so if you refer to a different _method_ in your docstring it won't be rewritten.

This is a fix for #332, which I opened.

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -309,7 +309,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
         // Methods
         for (m <- i.methods) {
           w.wl
-          writeDoc(w, m.doc)
+          writeMethodDoc(w, m, idCpp.local)
           val ret = marshal.returnType(m.ret, methodNamesInScope)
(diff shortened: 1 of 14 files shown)


artwyman:
Thanks for taking a stab at this.  Sounds good in abstract.  I haven't looked at the code in detail.

You should add some cases which trigger the new behavior in the test suite and/or example app, then re-run the generators so that we can see the generated output changes in this diff.

louiswins:
Will do

PR #354, opened 2018-02-21
Added support for constexpr for primitives in header files

alancast:
The problem here is that if you have a record of const ints, for example a record of error codes, you can not switch case them as the values are not set in the header. This PR adds constexpr support for const primitives in header files so that this is doable. It also fixes const enum values in Objective C, which appeared to never have been working.

It excludes optional primitives as they might not support constexpr (if using things like boost optional)

(It is based off PR https://github.com/dropbox/djinni/pull/282 which appears to have been abandoned)

* Got a basic version of constexpr in headers working

* got enum values to work as well

* can't do enums because interface headers forward declare so you can't set it there. So only primitives allowed

* fixing const enums in obj-c, which apparently never worked to begin with


xianwen:
Hi, Alan:

Thanks a lot!

alancast:
I _think_ I just signed it so we should be good to go now :-)

PR #355, opened 2018-03-06
Python - Node JS

egorafanasenko:
Is there any chance to work with djinni from Node JS? In same way as we work from Swift and Java.


artwyman:
@egorafanasenko.  I'm assuming you intended this as a question, not an attempt to merge the python branch into master, so I'm going to close out the PR.

To answer your question, I don't know of any reason Djinni's approach couldn't work for Node JS, and you could probably follow some of the same techniques used for the Python integration.  I don't know of anyone who's experimented with it, but if you wan to ask further you should join the Slack community mentioned in the README, where you'll find more developers who've worked with Djinni.

PR #365, opened 2018-05-11
Rename destroy() to __destroy()

sheldonneuberger:
This prevents compilation errors when someone creates a djinni interface with their own method named "destroy()".

### DIFF
--- a/src/source/JavaGenerator.scala
+++ b/src/source/JavaGenerator.scala
@@ -180,12 +180,12 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
             }
             w.wl
             w.wl("private native void nativeDestroy(long nativeRef);")
-            w.wl("public void __destroy()").braced {
+            w.wl("public void _djinni_destroy()").braced {
               w.wl("boolean destroyed = this.destroyed.getAndSet(true);")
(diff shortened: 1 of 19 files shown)


j4cbo:
Names with two leading underscores are reserved, so _destroy would be better (or, better yet, _djinni_destroy)

artwyman:
We have a use of a `djinni_private` prefix for a similar purpose here: https://github.com/dropbox/djinni/blob/master/support-lib/objc/DJIObjcWrapperCache%2BPrivate.h#L91

That being said, I don't have any serious objection to `_djinni_destroy`.  Only one underscore, and including the word `djinni` should make it plenty safe against conflicts, and consistency between the ObjC++ and JNI helpers isn't a big deal.

sheldonneuberger:
Tests pass. I changed it to _djinni_private_destroy so that it's more consistent with objc; doesn't hurt to have that. I prefer the underscore in front to make it more clear that it's special and private, but if someone objects we can also drop that leading underscore.

xianwen:
Hi, @sheldonneuberger:
Thanks a lot!

sheldonneuberger:
Signed.

sheldonneuberger:
@xianwen possible to merge this today?

PR #367, opened 2018-05-17
support-lib: don't let a noexcept function throw exceptions

libcg:
Fixes a compilation error following NDK update:

`../deps/djinni/support-lib/jni/djinni_support.cpp:456:9: error: 'jniDefaultSetPendingFromCurrent' has a non-throwing exception specification but can still
      throw [-Werror,-Wexceptions]
        throw;
        ^
../deps/djinni/support-lib/jni/djinni_support.cpp:453:6: note: function declared non-throwing here
void jniDefaultSetPendingFromCurrent(JNIEnv * env, const char * /*ctx*/) noexcept {
     ^                                                                   ~~~~~~~~`

### DIFF
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -594,10 +594,9 @@ void jniDefaultSetPendingFromCurrent(JNIEnv * env, const char * /*ctx*/) noexcep
         return;
     } catch (const std::exception & e) {
         env->ThrowNew(env->FindClass("java/lang/RuntimeException"), e.what());
+    } catch (...) {
+        std::terminate();
     }
-
-    // noexcept will call terminate() for anything not caught above (i.e.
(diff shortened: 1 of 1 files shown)


artwyman:
I've learned to be wary of `catch (...)` because it catches things you wouldn't think of as (C++) exceptions on some platforms.  My experience was on Windows, where it catches "Windows Structured Exception Handling" exceptions (an old C thing built on LONGJMP), as well as debugger breakpoints.

Per the comment after the catch, the intent of the code here is that any exception which isn't a std::exception should just call terminate() when it hits the noexcept spec, which the language guarantees.  That can probably be accomplished without the warning with another layer of indirection.  I.e. put the code inside a function with no exception spec, then call it from the main function marked noexcept.

artwyman:
Repalced by #378

Issue #377, opened 2018-06-18
`noexcept` in `jniDefaultSetPendingFromCurrent`


jonmcclung:
I get the following error when trying to run `make example_android`. I have made the following changes to `Application.mk`:
`NDK_TOOLCHAIN_VERSION := clang`
`APP_PLATFORM := android-23`
`APP_STL := c++_shared`
I am using the r17b NDK toolchain, because [GCC is deprecated and will be removed next iteration](https://developer.android.com/ndk/downloads/revision_history).

```
[arm64-v8a] Compile++      : djinni_jni <= djinni_support.cpp
../../../support-lib/jni/djinni_support.cpp:591:9: error: 'jniDefaultSetPendingFromCurrent' has a non-throwing exception specification but can still throw [-Werror,-Wexceptions]
        throw;
        ^
../../../support-lib/jni/djinni_support.cpp:588:6: note: function declared non-throwing here
void jniDefaultSetPendingFromCurrent(JNIEnv * env, const char * /*ctx*/) noexcept {
     ^                                                                   ~~~~~~~~
1 error generated.
```

It seems that the `jniDefaultSetPendingFromCurrent` function in `djinni_support.cpp` is marked `noexcept`, yet it still throws:
 
```
void jniDefaultSetPendingFromCurrent(JNIEnv * env, const char * /*ctx*/) noexcept {
    assert(env);
    try {
        throw;
    } catch (const jni_exception & e) {
        e.set_as_pending(env);
        return;
    } catch (const std::exception & e) {
        env->ThrowNew(env->FindClass("java/lang/RuntimeException"), e.what());
    }

    // noexcept will call terminate() for anything not caught above (i.e.
    // exceptions which aren't std::exception subclasses).
}
```

artwyman:
See discussion on PR #367

PR #378, opened 2018-06-18
Provides Clang Compatibility

jonmcclung:
Without this, clang will treat it as an error that `jniDefaultSetPendingFromCurrent` could potentially throw an uncaught exception. Compare PR #367 and issue #377. With this change, I had no issues compiling and running the Android example with clang from NDK r17b. 

Note: For anyone else trying to build with clang, I also made the following modifications locally, but did not commit them:

- remove everywhere the compiler flag -Wo-literal-suffix is used
- in `example/android/app/jni/Application.mk`:
  - `NDK_TOOLCHAIN_VERSION := clang`
  - `APP_PLATFORM := android-23`
  - `APP_STL := c++_shared`

### DIFF
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -601,8 +601,8 @@ void jniDefaultSetPendingFromCurrent(JNIEnv * env, const char * /*ctx*/) noexcep
 	jniDefaultSetPendingFromCurrentClangCompatibility(env);
 
     /* noexcept will call terminate() for anything not caught in 
-	   jniDefaultSetPendingFromCurrent (i.e. exceptions which aren't
-	   std::exception subclasses). */
+	   jniDefaultSetPendingFromCurrentClangCompatibility (i.e. 
+	   exceptions which aren't std::exception subclasses). */
 }
 
 template class ProxyCache<JavaProxyCacheTraits>;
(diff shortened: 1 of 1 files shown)


support-lib/jni/djinni_support.cpp
```diff
@@ -585,7 +585,7 @@ void jniSetPendingFromCurrent(JNIEnv * env, const char * ctx) noexcept {
     jniDefaultSetPendingFromCurrent(env, ctx);
 }
 
-void jniDefaultSetPendingFromCurrent(JNIEnv * env, const char * /*ctx*/) noexcept {
+void jniDefaultSetPendingFromCurrentClangCompatibility(JNIEnv * env) {
```
artwyman: Can you give this a name which describes what it is, rather than referring to the clang limitation?  Just `jniDefaultSetPendingFromCurrentImpl` would be fine, or else something decriptive like `jniThrowToSetPending`

That being said, a comment describing why the extra layer is here would be welcome too, though I'd suggest it refer to the general problem (not throwing in a function marked as noexcept) rather than the clang warning itself.

artwyman:
I can confirm the CLA has already been signed (thanks @jonmcclung) so this can merge after code-review is complete.

jonmcclung:
I think that should be what you asked for. Let me know what you think about the changes I've made.

artwyman:
LGTM.  Flagging @xianwen to include this the next time he tests and merges.

jonmcclung:
@artwyman Sorry, I'm not used to open source stuff. This is actually my first PR. It should be fine now.

PR #382, opened 2018-06-24
Add Android Parcelable records documentation

4brunu:
This PR adds Android Parcelable records documentation that was missing in https://github.com/dropbox/djinni/pull/308


4brunu:
@artwyman @xianwen Could you please take a look at this PR? Thanks

PR #388, opened 2018-07-18
Make #379 compatible with python branch

xianwen:
After merging master branch to python branch, the python compile failed because of the update in #379.
So I made another commit 875bbe2 to update `PythonGenerator.scala`, so that it can make use of the updated Doc object


xianwen:
@4brunu please take a look see if anything should be adjusted, thanks.

Issue #176, opened 2015-12-17
Calling back into Java from other threads


jefftrull:
I ran into an interesting issue today with calling back Java interfaces from non JVM attached threads.

There's a common pattern in asynchronous programming where you have a pool of worker threads handling the results of previous requests as they arrive.  We are using a third-party library that does this; the workers are executing previously stored callbacks which are `std::shared_ptr<Callback>`, where Callback is an interface generated by Djinni.  Unfortunately the call hits the code in jniGetThreadEnv() which aborts if the JVM getEnv() call fails, which it will when a thread (such as the workers created by the library) is not attached to a JVM already.  We worked around it like this:

```
-    const jint get_res = g_cachedJVM->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION_1_6);
+    jint get_res = g_cachedJVM->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION_1_6);
+    if (get_res == JNI_EDETACHED) {
+        // attach this thread to the JVM
+        get_res = g_cachedJVM->AttachCurrentThread(&env, nullptr);
+    }
```

but it seems a little hacky.  Any thoughts on a better approach?  Thanks!

artwyman:
We avoid C++-created threads calling into Java because that situation can call other issues beyond thread attach.  Specifically the thread may not use the right class loader, such that calls into Java might fail to find a class they need.  I think there might also be a risk of resource leakage when attached threads terminate, but I'm not sure of that.

I can't fully describe exactly what the issues are and how to avoid them, since we decided early on to avoid the situation entirely.  There's some sample code in djinni/extension_libs for the platform threads interface we use to let C++ create threads by calling into Java.  Of course, your 3rd-party library might not let you control thread creation, in which case you'd need to either post to another thread which was created by Java, or do the attach and deal with any fallout (pre-loading classes might be enough).

If you're going to do the attaching, then your approach seems reasonable.  The other alternative I'd see would be to just always call AttachCurrentThread, which should be a nop on an already-attached thread, and will give you the env.  I'm not sure I'd want to make this approach part of the mainline Djinni code without opt-in configurability, and/or some way to deal with detaching threads.

jefftrull:
OK, thanks for the advice!

Issue #372, opened 2018-06-04
Provide a way to attach a c++ thread to the JVM


sonicdebris:
In our application, we are calling java listeners from a background thread, and we faced issues similar to #238, #219, #176 and #141 . As far as I understand, the main concern for not using `AttachCurrentThread` in Djinni is because of uncertainty about when to call `DetachCurrentThread`.

Actually, as adviced on the NDK website's [Jni Tips](https://developer.android.com/training/articles/perf-jni#threads) page, it would be possible to reliably detach the thread when it exits.

I've already used this mechanism in a few big android projects and it's working quite well. One needs to define a thread exit callback, create a `pthread_key` once per process (a good place could be where the jvm is cached)
`
static pthread_key_t threadKey;

void onThreadExit(void*)
{
    jvm->DetachCurrentThread();
}

void createThreadKey()
{
    pthread_key_create(&threadKey, onThreadExit);
}
`
and then lazily attach the thread to the jvm and associate the key to the thread itself, the first time the env is needed for that thread, something like:
`
JNIEnv* getEnv()
{
    JNIEnv* env;
    jint res = jvm->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION_1_6);

    if (res == JNI_EDETACHED)
    {
        res = jvm->AttachCurrentThread(&env, nullptr);
        assert(res == JNI_OK);
        pthread_setspecific(threadKey, env);
    }

    return env;
}
`
For now we are calling this manually when creating a thread where we know java methods will be called using djinni, and it's working fine, so maybe we could try integrating this logic directly in djinni's `jniGetThreadEnv` (so we can avoid caching the jvm ourselves, too, and having to stub the function for non-android platforms). 

If you are interested, I could submit a pull request with the changes.

artwyman:
That sounds like a reasonable approach.  It does seem to make some assumptions about the platform, though, and I'm not educated enough on pthreads to know exactly what the safe vs. unsafe cases would be.  E.g. I'm not sure whether threads created in ways other than pthreads are guaranteed to call pthread_exit to do this kind of cleanup.  Also use of pthreads at all would seem to rule out Windows (or anything non-POSIX) as a platform.  Android is indeed the primary target platform for Djinni's Java support, but it's meant to generalize to any Java platform (the unit tests run on OSX or Linux for instance).

[...]

PR #398, opened 2018-08-20
Provide a method to run functions on JVM thread

jonmcclung:
This is an attempt to provide a way for C++-created threads to call into Java without risking a memory leak, as requested in #176 . The basic idea is this:

- you pass in a lambda
- the current thread is attached to the JVM if it's not already
- the lambda is executed and the result, if any, stored
- the current thread is detached only if it was this function that attached it
- the result, if it was stored, is returned.

There are two functions because I'm not skilled enough with lambdas to make one function that will accept any kind of lambda.

### DIFF
--- a/support-lib/jni/djinni_support.hpp
+++ b/support-lib/jni/djinni_support.hpp
@@ -47,6 +47,48 @@ void jniShutdown();
  */
 JNIEnv * jniGetThreadEnv();
 
+JavaVM * getCachedJVM();
+/*
+ * Runs the given function on a Java thread and subsequently detaches it if it was not
+ * previously attached.
+ */
+template<typename Func>
+typename std::result_of<Func()>::type runOnJvmThreadWithReturn(Func f) {
+    auto g_cachedJVM = getCachedJVM();
+    assert(g_cachedJVM);
+    bool attached = false;
+    JNIEnv * env = nullptr;
+    jint get_res = g_cachedJVM->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION_1_6);
+    if (get_res == JNI_EDETACHED) {
+        get_res = g_cachedJVM->AttachCurrentThread(&env, nullptr);
+        attached = true;
+    }
+    if (get_res != 0 || !env) {
+        std::abort();
+    }
+    auto result = f();
+    if (attached) g_cachedJVM->DetachCurrentThread();
+    return result;
+}
+
+template<typename Func>
+void runOnJvmThread(Func f) {
+    auto g_cachedJVM = getCachedJVM();
+    assert(g_cachedJVM);
+    bool attached = false;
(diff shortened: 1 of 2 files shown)


mjmacleod:
I've tested this and confirm it works my side.

However I don't personally like the extra code clutter to use this, nor the overhead of constantly attaching/detaching.
I've implemented the code suggestions in #372 as an alternative pull request #405 that could possibly be used instead of this though I guess there is no reason they could not both co-exist also.

support-lib/jni/djinni_support.hpp
```diff
@@ -47,6 +47,48 @@ void jniShutdown();
...
+    jint get_res = g_cachedJVM->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION_1_6);
+    if (get_res == JNI_EDETACHED) {
+        get_res = g_cachedJVM->AttachCurrentThread(&env, nullptr);
+        attached = true;
+    }
+    if (get_res != 0 || !env) {
+        std::abort();
+    }
+    f();
+    if (attached) g_cachedJVM->DetachCurrentThread();
```
shahzadlone: Please make the if one line if's consistent with the rest, just have a {} for if's scope like the majority.
```diff
-    if (attached) g_cachedJVM->DetachCurrentThread();
+    if (attached) {
+        g_cachedJVM->DetachCurrentThread();
+    }
```

support-lib/jni/djinni_support.hpp
```diff
@@ -47,6 +47,48 @@ void jniShutdown();
...
+    jint get_res = g_cachedJVM->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION_1_6);
+    if (get_res == JNI_EDETACHED) {
+        get_res = g_cachedJVM->AttachCurrentThread(&env, nullptr);
+        attached = true;
+    }
+    if (get_res != 0 || !env) {
+        std::abort();
+    }
+    auto result = f();
+    if (attached) g_cachedJVM->DetachCurrentThread();
```
shahzadlone: Please make the if one line if's consistent with the rest, just have a {} for if's scope like the majority.
```diff
-    if (attached) g_cachedJVM->DetachCurrentThread();
+    if (attached) {
+        g_cachedJVM->DetachCurrentThread();
+    }
```

mjmacleod:
Just a note that #405 has been merged now, I'm not sure if there is a good reason to have more than one way to do this or not.

shahzadlone:
> Just a note that #405 has been merged now, I'm not sure if there is a good reason to have more than one way to do this or not.

That's fair enough. I just pointed out because I have seen some pretty tedious bugs introduced because of the missing braces after an if.

artwyman:
#405 provided an alternative option to this.  This repo isn't going to be adding any new features so I'm putting this PR on hold.

PR #400, opened 2018-09-07
Fix misleading README instructions

mjmacleod:
README instructions do not generate the files they say they will generate, fix the djinni command called so that they match up.


mjmacleod:
Replaced by #411

Issue #390, opened 2018-07-27
Request: allow option to choose the type of list in Java


jdevera:
`list<Something>` in IDL generates `ArrayList<Something>` in Java, but this makes for too restrictive APIs.

I would like to find a way to map `list` to `List` in method parameter types.

How can I do this?

Probably the same applies for `set`.

PR #406, opened 2018-09-19
Change djinni's generated Java files to use interfaces

ggdabuh:
This pull request is related to https://github.com/dropbox/djinni/issues/390.
The idea is to use interfaces rather that classes as, anyway, the marshaling layer is ready for that.
This is a breaking change since users will need to update their overrides.

### DIFF
--- a/src/source/JNIMarshal.scala
+++ b/src/source/JNIMarshal.scala
@@ -70,9 +70,9 @@ class JNIMarshal(spec: Spec) extends Marshal(spec) {
         case MOptional => throw new AssertionError("nested optional?")
         case m => javaTypeSignature(tm.args.head)
       }
-      case MList => "Ljava/util/ArrayList;"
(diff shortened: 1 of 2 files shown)


artwyman:
To be clear for readers, is about using abstract interfaces for generated container types like list and map when used in records, arguments, and returns.  This isn't about what Djinni calls interfaces.

A few procedural things:
- You should re-generate the test suite and example code so we can see how your generator change affects the generated code for these examples.  You should also ensure the test suite passes (if you can `make all` from the root of the repo you're all set).
- This would be a breaking change for existing users, so I think it needs to be exposed as a configurable feature via command-line, not a default behavior change

xianwen:
Hi, @ggdabuh:
Thanks a lot!

artwyman:
This repo isn't going to be adding any new features so I'm putting this PR on hold.

PR #413, opened 2018-10-26
Add cast to objective C convenience initializers to prevent naming co…

wpurdy2:
In generated Objective C code, convenience initializers can easily cause naming collisions with built in ios functions and other code that follows the same naming pattern.  As a trivial example, consider the following record:

`MyColor = record
{
    red : f32;
    green : f32;
    blue : f32;
    alpha : f32;
} deriving( eq )`

This produces a function that looks something like this:
` (nonnull instancetype)MyColorWithRed:(float)red
{
(nonnull instancetype)MyColor:(float)red
                                   green:(float)green
                                    blue:(float)blue
                                   alpha:(float)alpha
{
    return [[self alloc] initWithRed:red
                                                green:green
                                                 blue:blue
                                                alpha:alpha];
}
}`

If your project happens to include UIKit, this generated code will emit a compilation error because InitWithRed is defined in UIKit (see https://developer.apple.com/documentation/uikit/uicolor/1621925-initwithred?language=objc) and without qualification, the call is ambiguous.

Adding a cast to the generated code disambiguates the call to initWithRed (and other similar situations)

### DIFF
--- a/src/source/ObjcGenerator.scala
+++ b/src/source/ObjcGenerator.scala
@@ -235,7 +235,7 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
         writeAlignedObjcCall(w, decl, r.fields, "", f => (idObjc.field(f.ident), s"(${marshal.paramType(f.ty)})${idObjc.local(f.ident)}"))
         w.wl
         w.braced {
-          val call = s"return [[self alloc] init$firstInitializerArg"
+          val call = s"return [($self*)[self alloc] init$firstInitializerArg"
           writeAlignedObjcCall(w, call, r.fields, "", f => (idObjc.field(f.ident), s"${idObjc.local(f.ident)}"))
           w.wl("];")
         }
--- a/example/generated-src/objc/TXSItemList.mm
+++ b/example/generated-src/objc/TXSItemList.mm
@@ -16,7 +16,7 @@
 
 + (nonnull instancetype)itemListWithItems:(nonnull NSArray<NSString *> *)items
 {
-    return [[self alloc] initWithItems:items];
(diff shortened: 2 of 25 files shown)


src/source/ObjcGenerator.scala
```diff
@@ -235,7 +235,7 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
         writeAlignedObjcCall(w, decl, r.fields, "", f => (idObjc.field(f.ident), s"(${marshal.paramType(f.ty)})${idObjc.local(f.ident)}"))
         w.wl
         w.braced {
-          val call = s"return [[self alloc] init$firstInitializerArg"
+          val call = s"return [($self*)[self alloc] init$firstInitializerArg"
```
artwyman: Would a cast to instancetype make more sense here?  Looking at the generated code it seems inconsistent to use the explict type here but `instancetype` in the return type declaration 2 lines above.  Of course, I'm a bit surprised by the ObjC behavior you're fixing here too, so I'm not sure if it's more quirky than I expect.

wpurdy2: To be honest, I have only worked on Objective C code a handful of times so I may be off here, but my understanding of the instancetype keyword is that it can only be used as a return type in function signatures, and casting to it is not possible.  Doing [(instancetype)[self alloc] results in a compiler error, as does instancetype* .

artwyman: Huh, I was always a polyglot too so that may be a quirk or Objective C I'm not clear on.  It would be nice if instancetype could be used more broadly, but if it doesn't then the direct cast is the best we can do.

xianwen:
Hi, @wpurdy2:
Thanks a lot!

PR #414, opened 2018-10-30
Add the ability to get the JavaVM

csabaszigetipix4d:
Add the ability to get the JavaVM which was registered in the onload for e.g. forwarding to other libraries

### DIFF
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -69,6 +69,10 @@ void jniShutdown() {
     g_cachedJVM = nullptr;
 }
 
+JavaVM * jniGetVM() {
+    return g_cachedJVM;
+}
+
 JNIEnv * jniGetThreadEnv() {
     assert(g_cachedJVM);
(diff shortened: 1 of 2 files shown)


support-lib/jni/djinni_support.cpp
```diff
@@ -594,6 +598,8 @@ void jniDefaultSetPendingFromCurrentImpl(JNIEnv * env) {
         return;
     } catch (const std::exception & e) {
         env->ThrowNew(env->FindClass("java/lang/RuntimeException"), e.what());
+    } catch(...) {
```
artwyman: This doesn't look related to the purpose of this PR.  Can you explain?
I've generally learned to shy away from `catch(...)` since depending on the platform/compiler it can catch things you didn't intend, including things which aren't really C++ exceptions.

csabaszigetipix4d: Sorry, it is really not related. I revert this change now.

xianwen:
Hi, @csabaszigetipix4d:
Thanks a lot!

PR #415, opened 2018-10-30
Add support for `-fvisibility=hidden` and `-fvisibility=default` flags

csabaszigetipix4d:
I quote the updated README.md here : 

## Support for `-fvisibility=hidden` and `-fvisibility=default`

Djinni supports both the `-fvisibility=hidden` and the `-fvisibility=default` flags.

The symbols that are belonging to the public interfaces that are generated by Djinni are automatically defined to be visible symbols.
To achieve this Djinni is marking the generated code with `PROJECT_EXPORT`.
`PROJECT_EXPORT` is defined in `support-lib/djinni_common.hpp`.
This macro is using the correct attribute specifier for the actual compiler. For more details, please check the macro definition in the `support-lib/djinni_common.hpp` header.
This `PROJECT_EXPORT`  does nothing in case if you are using the `-fvisibility=default` compiler flag.
This `PROJECT_EXPORT` in case if you are using the `-fvisibility=default` compiler flag.

Since the generated headers are including this `support-lib/djinni_common.hpp`, you must distribute it together with your library that is using Djinni.

### Windows

For Windows builds, you must define `BUILDING_DLL` based on your needs.
You can define it as a compiler option like `-DBUILDING_DLL`.
For more details, please check the macro definition in the `support-lib/djinni_common.hpp` header.


xianwen:
Hi, @csabaszigetipix4d:
Thanks a lot!

CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ include(GNUInstallDirs)
 
 set(SRC_SHARED
   "support-lib/djinni_common.hpp"
+  "support-lib/project_export.hpp"
```
artwyman: This is used so widely (requiring it to be in the include path) that it really needs a more unique name, like `djinni_project_export.hpp`.

README.md
```diff
@@ -515,6 +515,23 @@ In order to do that, there are two steps needed:
 - deriving the records that should be parcelable with the keyword parcelable: `deriving(parcelable)`
 - run Djinni with the following flag `--java-implement-android-os-parcelable true`
 
+## Support for `-fvisibility=hidden` and `-fvisibility=default`
+
+You can pass `-fvisibility=hidden` or the `-fvisibility=default` flags to your compiler, Djinni handles both cases well.
+
+The symbols that are belonging to the public interfaces that are generated by Djinni are automatically defined to be visible symbols.
```
artwyman: I feel like it's not obvious that making these symbols always public is the right behavior for all users.  E.g. if Djinni symbols are only used internally to a larger library, they wouldn't need to be exported.  In the use case we've used at Dropbox, we build all our C++ code into a single library on Android, so only the JNI functions need to be exported.

PR #427, opened 2019-02-02
static libraries should set SKIP_INSTALL=YES in XCode builds

adamtait:
Every time I try to generate an iOS App archive from XCode, I run into the problem that XCode doesn't recognize the archive as an iOS App because SKIPS_INSTALL isn't set of the Djinni support lib targets. [Here's an Apple technical note explaining](https://developer.apple.com/library/archive/technotes/tn2215/_index.html).

I think the Djinni support lib targets should always be static libraries, so I think this is the right behavior to have by default. Also, I'm hoping that others won't be tripped up like I was since XCode provides no feedback in debugging why an archive isn't considered an iOS App.

### DIFF
--- a/extension-libs/platform-threads/platform-threads.gyp
+++ b/extension-libs/platform-threads/platform-threads.gyp
@@ -14,12 +14,16 @@
                   "cpp",
                 ],
             },
+            "xcode_settings": {
+              "SKIP_INSTALL": "YES",
+            },
         },
         {
             "target_name": "thread_objc",
             "type": "static_library",
             "xcode_settings": {
               "CLANG_ENABLE_OBJC_ARC": "YES",
+              "SKIP_INSTALL": "YES",
             },
             "sources": [
               "objc/DJIThreadFactoryImpl.h",
--- a/support-lib/support_lib.gyp
+++ b/support-lib/support_lib.gyp
@@ -17,6 +17,9 @@
           "jni",
         ],
       },
+      "xcode_settings": {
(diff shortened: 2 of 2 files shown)


adamtait:
I've just signed the [Dropbox CLA](https://opensource.dropbox.com/cla).

adamtait:
> The typical build procedure I've seen before (and which we've been using at Dropbox for years) is to build a bunch of static libraries, and then link them all together into the app binary, or a dynamic library for inclusion in the app. In that case it doesn't seem like this setting is needed.

This is the same flow that I'm describing. I am including Djinni's support libs as static library targets in another target app binary. 

In order to build an iOS App archive, XCode requires that all targets set `"SKIP_INSTALL": "NO"`. By default, XCode assumes "YES". The result is that you aren't able to produce an iOS App archive; since you have more than one bundled target that can be "installed", XCode will not consider your bundle an iOS App (and you can't get it into the App Store). This has tripped me up many times now.

> If you do need that setting in your particular use case, one thing you might consider is adding it in your common.gypi as a default setting for all targets, then only override it to NO on your app targets.

I hadn't thought of this solution until now, just tested and can confirm that it also works. [I've also just noticed that `SKIP_INSTALL` is set correctly to `YES` in the common.gypi included in Djinni](https://github.com/dropbox/djinni/blob/master/common.gypi#L14). I'm not using the included common.gypi but this seems like a valid and preferred option, so I'm going to close this PR.

PR #433, opened 2019-03-07
Fix build Android example project

4brunu:
The Android example project it's not building anymore due to multiple incompatibilidades with the last Android Studio release 3.3.2 as today.
So I fix all the issues, because this was not letting me run the command `make all` with success.
Where it was possible I created a comment near the issue, showing the error, for future documentation on why things changed.

### DIFF
--- a/common.gypi
+++ b/common.gypi
@@ -4,7 +4,7 @@
     # enable android short names (not full paths) for linking libraries
     "android_unmangled_name": 1,
     'cflags':    [ '-gdwarf-2', '-Werror', '-Wall', '-Wextra', '-Wno-missing-field-initializers' ],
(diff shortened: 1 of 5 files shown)


4brunu:
Thanks

artwyman:
Are these issues are really tied to the updated Android Studio version or NDK version?  I was aware of the NDK's move to clang, though I'm not sure if I want to take on the responsibility of making that update at the risk of breaking changes.

4brunu:
@artwyman some changes are related to Android Studio and some to NDK, but the most relevant like clang are related to NDK.
I created this update, because someone that has the last version of Android Studio and NDK needs those changes to run the tests.

artwyman:
I agree that updating to the latest version will be the most useful for more users.  I'm going to hold off until I can test this against both old and new NDKs to understand the impact there, but plan to get it merged if I can work that out.

Issue #408, opened 2018-10-06
Android Parcelable missing ArrayList import


skonstant:
I have a record that has sets no lists, and when the parcelable code is generated it creates something like:

 ArrayList<Long> mSubscriptionoasisTemp = new ArrayList<Long>();
 in.readList(mSubscriptionoasisTemp, getClass().getClassLoader());
 this.mSubscriptionoasis = new HashSet<Long>(mSubscriptionoasisTemp);

But since there is no list in the record, the import directive for ArrayList is not there, which breaks compilation.

The fix is not simple, I have a quick workaround for now but I don't think it will work with records that do have lists, it might import twice and break compilation again.

In JavaGenerator.scala:

override def generateRecord(origin: String, ident: Ident, doc: Doc, params: Seq[TypeParam], r: Record) {
    val refs = new JavaRefs()
    r.fields.foreach(f => refs.find(f.ty))

    if (spec.javaImplementAndroidOsParcelable && r.derivingTypes.contains(DerivingType.AndroidParcelable))
      refs.java += "java.util.ArrayList"

The last two lines fix it.

artwyman:
@skonstant want to submit a PR for this fix?

I also wonder if the simple workaround you have above might be too broad for cases where a record implements Parcelable but doesn't contain any container types at all (sets, maps, lists) so won't need any readList() calls at all.

4brunu:
I just created a PR https://github.com/dropbox/djinni/pull/434 that fixes this issue.

prsolucoes:
@artwyman can you merge it?

PR #434, opened 2019-03-07
Fix an issue with parcelable code gen with sets and no lists

4brunu:
This PR fixes https://github.com/dropbox/djinni/issues/408.
I added some conditions to only add the import of `ArrayList` when necessary, avoiding the import duplication.


4brunu:
Thanks

src/source/JavaGenerator.scala
```diff
@@ -249,6 +249,35 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
     val javaName = if (r.ext.java) (ident.name + "_base") else ident.name
     val javaFinal = if (!r.ext.java && spec.javaUseFinalForRecord) "final " else ""
 
+    def checkIfRecordContainsSets(r: Record): Boolean = {
```
artwyman: nit: I'd prefer that predicates like this are named as a question or statement.  E.g. recordContainsSets() or doesRecordContainSets().

src/source/JavaGenerator.scala
```diff
@@ -249,6 +249,35 @@ class JavaGenerator(spec: Spec) extends Generator(spec) {
...
+      for (f <- r.fields) {
+        f.ty.resolved.base match {
```
artwyman: This fix addresses the specific test case, but doesn't go deep enough.  E.g. if the record contains an optional<set<string>> this function won't find it.  I tried such a record and was able to see the Java compile fail.  I think you need to recurse into the args as well as looking at the base type.

4brunu: "I think you need to recurse into the args as well as looking at the base type."
Can you elaborate a bit more on this?

artwyman: `r.args` contains the type args.  So for instance with `optional<set<string>>` `r.base` would be `MOptional`, while args would contain one item whose `base` would be `MSet`.  I think you need to search for `MSet` in the args recursively to find cases like that, or cases like `list<set<string>>`, etc.

4brunu: Done 👍

4brunu:
@artwyman all the issues are now fixed

src/source/JavaGenerator.scala
artwyman: I still don't think this is sufficient.  I was using optional<> as an example, not the only relevant type.  You'd have the same problem with map<string, set<string>>, for instance.
I don't think you need to hard-code any specific set of types.  Just always recurse by calling this same function on each element of `f.ty.resolved.args`, and combine the results with logical or.

4brunu: @artwyman The parcelable generated code doesn't need to deal with nested collections.
`Optional<>` was a corner case that I didn't remembered.
But a `set<strings>` is the same as a `set<list<string>>` or a `set<map<string, string>>`, and the same applies to the other collections, the `optional<>` was the only exception.
Check the unit tests to see what I'm referring to.

4brunu:
@artwyman I also added more unit tests to the parcelable feature to ensure more code coverage to the feature.

src/source/JavaGenerator.scala
artwyman: Oh, I see, what's needed here isn't a full recursive processing of the type but only one level.  I assumed recursive, and took the `optional<set<string>>` case as an indication that theory held true.  I just thought about `optional<optional<set<string>>>` now, but we explicitly disallow direct nesting of optionals.

It's surprising to that `writeToParcel` of `map<string, set<string>>` doesn't need to use `writeList` in the same way that it does for `set<string>`.  But Thanks for adding the additional test cases to demonstrate that all of those cases work.

Issue #431, opened 2019-02-25
Djinni don't write enums correctly using ordinal value


prsolucoes:
Hi,

Djinni don't write on parcel enum value as ordinal. In my example:

```
/*package*/ final com.ubook.library.core.enums.SearchParamOrderAscendingTypeEnum mOrderAscensing;

    @Override
    public void writeToParcel(android.os.Parcel out, int flags) {
        out.writeList(this.mCatalogTypeList);
        out.writeString(this.mOrderField);
        out.writeInt((int)this.mOrderAscensing); // error when compile
        out.writeString(this.mSource);
        out.writeInt((int)this.mDownloaded);  // error when compile
        out.writeString(this.mText);
    }
```

Java generate enum:

```
public enum SearchParamOrderAscendingTypeEnum {
    IGNORE,
    ASCENDING,
    DESCENDING,
    ;
}
```

I fix it manually using:

```
    @Override
    public void writeToParcel(android.os.Parcel out, int flags) {
        out.writeList(this.mCatalogTypeList);
        out.writeString(this.mOrderField);
        out.writeInt((int)this.mOrderAscensing.ordinal());
        out.writeString(this.mSource);
        out.writeInt((int)this.mDownloaded.ordinal());
        out.writeString(this.mText);
    }
```

Can you fix it please?

Thanks.

artwyman:
This feature was contributed by @4brunu in #308, so perhaps he can take a look.

4brunu:
Hum, maybe I didn't test for this case, @prsolucoes do you want to try to fix it and create a PR? I'm a bit busy, but I can try to fix it once I have time.

4brunu:
@prsolucoes Are you using the last version of djinni, from master branch?
I was testing this, and it's working in my test case

```
    public ItemList(android.os.Parcel in) {
        this.mOrder = SortOrder.values()[in.readInt()];
    }

    @Override
    public void writeToParcel(android.os.Parcel out, int flags) {
        out.writeInt(this.mOrder.ordinal());
    }
```
Can you share a djinni example file?
Thanks

prsolucoes:
HI,

Yes, im using latest version:
![image](https://user-images.githubusercontent.com/395096/53898849-a8af5880-4017-11e9-840d-d7c1a32c24da.png)

Djinni file:

```
search_param_downloaded_type_enum = enum {
    ignore;
    downloaded;
    not_downloaded;
}

search_param_order_ascending_type_enum = enum {
[...]

PR #435, opened 2019-03-07
Fix some issues with parcelable code gen with enums and flags

4brunu:
The Flag type was added, but the support for parcelable for it was missing, so this PR add's it.
This also fixes https://github.com/dropbox/djinni/issues/431 and improves the test coverage for parcelable.

### DIFF
--- a/example/example.djinni
+++ b/example/example.djinni
@@ -1,6 +1,6 @@
 item_list = record {
     items: list<string>;
-}
+} deriving(parcelable)
 
 sort_order = enum {
     ascending;
--- a/example/example.djinni
+++ b/example/example.djinni
@@ -1,6 +1,6 @@
 item_list = record {
     items: list<string>;
-} deriving(parcelable)
+}
 
 sort_order = enum {
     ascending;
--- a/example/run_djinni.sh
+++ b/example/run_djinni.sh
@@ -55,6 +55,7 @@ fi
     --java-package $java_package \
     --java-class-access-modifier "package" \
     --java-generate-interfaces true \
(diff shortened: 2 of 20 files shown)


4brunu:
Thanks

example/example.djinni
```diff
@@ -1,6 +1,6 @@
 item_list = record {
     items: list<string>;
-}
+} deriving(parcelable)
```
artwyman: Why is this needed in the example?  Seems unnecessary, and the example should stay simple and understandable.
Given the generated code isn't updated either, I suspect maybe this was a test case you didn't intend to push.

example/run_djinni.sh
```diff
@@ -55,6 +55,7 @@ fi
     --java-package $java_package \
     --java-class-access-modifier "package" \
     --java-generate-interfaces true \
+    --java-implement-android-os-parcelable true \
```
artwyman: As above, I think this feature shouldn't be included in the basic example.

4brunu: I included this here, because this way, if someone is developing a new feature in djinni and that feature breaks parcelable, it will notice faster, but if you want, I can remove it.

artwyman: A good goal, though I think the test-suite covers tat, without the example having to be involved.  As-is, the example doesn't build for non-Android Java, which is one of the included build flavors.

prsolucoes:
It will be merged too?

4brunu:
@artwyman All your suggestions are done 👍

PR #436, opened 2019-03-18
Updating be able to use newer Android NDK versions

arlm:
This patch upgrades the minimum for building using Android NDK 19:

* Upgrading the APP_PLATFORM to `android-16` (`Android NDK: android-8 is unsupported. Using minimum supported version android-16. `)
* Changing APP_STL from `gnustl_static` to `c++_static` (`Android NDK: APP_STL gnustl_static is no longer supported. Please switch to either c++_static or c++_shared. See https://developer.android.com/ndk/guides/cpp-support.html for more information.`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ intellij-plugin/djinni.jar
 
 # profiling output
 callgrind.out.*
+.project
+example/android/.settings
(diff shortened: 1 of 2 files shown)


artwyman:
Thanks for submitting, @arlm.  There's already a similar PR #433, so I'm closing this as a duplicate.

PR #437, opened 2019-03-26
Generated closed enums to better interop with Swift

4brunu:
This PR intents to enhance the Swift 5 and generated objc interoperability in Enums.

Before Swift 5, all C enums were closed, but in Swift 5 they become open by default in an effort to prepare Swift for binary compatibility.

This is useful for third party code, but since djinni is used to generate the bridge between objc and c++ inside an app and all the code, written on generated are controlled inside the teams responsible by them, all enums should be closed by default.

https://github.com/apple/swift-evolution/blob/master/proposals/0192-non-exhaustive-enums.md

**EDIT**: IF you prefer I can introduce a new flag, to choose between closed or open enums, although I think the closed enums will almost always be the preferred option, since it's the default Swift enums option, and Swift has safety features that works best with closed enums.

### DIFF
--- a/example/generated-src/objc/TXSSortOrder.h
+++ b/example/generated-src/objc/TXSSortOrder.h
@@ -8,4 +8,4 @@ typedef NS_ENUM(NSInteger, TXSSortOrder)
     TXSSortOrderAscending,
     TXSSortOrderDescending,
     TXSSortOrderRandom,
-};
+}__attribute__((enum_extensibility(closed)));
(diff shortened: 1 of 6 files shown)


artwyman:
Closed vs. open does seem like it would be a developer choice in theory, based on whether they need to support future binary-compatible extensions of their API.  I agree that most devs will want closed by default.

Regardless, though, this repo isn't going to be adding any new features so I'm putting this PR on hold.

4brunu:
What if we introduce a new flag, to choose between closed or open enums?

artwyman:
I think a new cmdline flag is a good workaround for potential breaking changes, so I'd be willing to merge that.
This might even be the kind of thing which could be represented by a new sort of `deriving`, but that might be overkill

4brunu:
@artwyman I added a command line argument for avoiding breaking changes.

PR #438, opened 2019-03-26
Fix example objc project

4brunu:
Xcode 10 doesn't support Swift 3 anymore, so I updated the example project to build again.


artwyman:
Thanks for addressing this, though I think it might be too early to merge it.  Am I right to understand that this issue is specific to Xcode 10.2?  I'm able to build the example project with Xcode 10.0 just fine.

I think this PR also makes Xcode 10.2 the minimum requirement, by setting Swift version to 5.  I think it may be too early to require that as a minimum.  Dropbox is holding at 10.1 for a little while due to the OS version requirements affecting test environments.  Maybe it would make sense to merge an alternative PR which increased the Swift version but not all the way to 3.

4brunu:
@artwyman I just updated this PR, to only update the Swift version to 4.2, which is compatible with Xcode 10.0

PR #452, opened 2019-10-07
added generated function enum_to_string

ctin:
I extended c++ generator by adding string representation of enum values.
Rules:
* now generated enum will have cpp file with a function body;
* serialize function named enum_to_string to avoid conflicts with std::enum;
* serialize function moved out from namespace to allow user use it in templates or factories;
* function name depends on user style. In sneak case it will be `enum_to_string`, but in camel case it will be `enumToString`
* function returns std::string or std::wstring depending on djinni option `cpp-use-wide-strings`

tests passed
test makefile changed to support Xcode 11

note: we should ignore temp output :(
note2: writeCppFileGeneric does not support writing code outside of namespace, I added that feature. The reason written above: some functions should not be wrapped by a namespace


artwyman:
Seems reasonable, but this repo isn't taking any new feature enhancements, per the note in the README:

> Maintenance note: This repo is stable but no longer actively maintained. No further development is planned, and no new feature PRs will be merged. Bug fixes may be merged on a volunteer basis. If you have questions or want to talk to other users of Djinni, you can join the Slack community via the link at the end of this document.

PR #453, opened 2019-10-18
Fix old cast style use in generated file and support lib.

canatella:
Fixes LLVM warning: use of old-style cast [-Wold-style-cast]

### DIFF
--- a/src/source/JNIGenerator.scala
+++ b/src/source/JNIGenerator.scala
@@ -173,7 +173,7 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
           w.wl(s"const auto& data = ::djinni::JniClass<$jniHelper>::get();")
         writeAlignedCall(w, "return {", r.fields, "}", f => {
           val fieldId = "data.field_" + idJava.field(f.ident)
-          val jniFieldAccess = toJniCall(f.ty, (jt: String) => s"jniEnv->Get${jt}Field(j, $fieldId)")
+          val jniFieldAccess = toJniCall(f.ty, (jt: String) => s"jniEnv->Get${jt}Field(j, $fieldId)") + ")"
           jniMarshal.toCpp(f.ty, jniFieldAccess)
         })
         w.wl(";")
@@ -298,7 +298,9 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
             }
             else
               w.w(")")
-            w.wl(";")
+            if (m.ret.nonEmpty)
+              w.w(")");
+            w.wl(";");
             w.wl(s"::djinni::jniExceptionCheck(jniEnv);")
             m.ret.fold()(ty => {
               (spec.cppNnCheckExpression, isInterface(ty.resolved)) match {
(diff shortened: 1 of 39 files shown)


artwyman:
Can you please re-generate the code for the example and test cases to show the impact of the generator changes?  Also please ensure the tests build and pass.  You can run `make all` from the root which will do everything, if you're on a Mac).

canatella:
There you go! Test are passing now and generated code looks ok.

test-suite/generated-src/jni/NativeAssortedPrimitives.cpp
```diff
@@ -35,20 +35,20 @@ auto NativeAssortedPrimitives::toCpp(JNIEnv* jniEnv, JniType j) -> CppType {
...
-            ::djinni::F64::toCpp(jniEnv, jniEnv->GetDoubleField(j, data.field_mFsixtyfour)),
-            ::djinni::Optional<std::experimental::optional, ::djinni::Bool>::toCpp(jniEnv, jniEnv->GetObjectField(j, data.field_mOB)),
-            ::djinni::Optional<std::experimental::optional, ::djinni::I8>::toCpp(jniEnv, jniEnv->GetObjectField(j, data.field_mOEight)),
-            ::djinni::Optional<std::experimental::optional, ::djinni::I16>::toCpp(jniEnv, jniEnv->GetObjectField(j, data.field_mOSixteen)),
-            ::djinni::Optional<std::experimental::optional, ::djinni::I32>::toCpp(jniEnv, jniEnv->GetObjectField(j, data.field_mOThirtytwo)),
-            ::djinni::Optional<std::experimental::optional, ::djinni::I64>::toCpp(jniEnv, jniEnv->GetObjectField(j, data.field_mOSixtyfour)),
-            ::djinni::Optional<std::experimental::optional, ::djinni::F32>::toCpp(jniEnv, jniEnv->GetObjectField(j, data.field_mOFthirtytwo)),
-            ::djinni::Optional<std::experimental::optional, ::djinni::F64>::toCpp(jniEnv, jniEnv->GetObjectField(j, data.field_mOFsixtyfour))};
+    return {::djinni::Bool::toCpp(jniEnv, (jniEnv->GetBooleanField(j, data.field_mB))),
```
artwyman: This seems to be generating a lot of extra parens, as in this line.  Can you make that conditional so they're only generated when necessary as part of generating a cast expression, not always?

src/source/JNIGenerator.scala
```diff
@@ -404,11 +406,11 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
 
   def toJniCall(ty: TypeRef, f: String => String): String = toJniCall(ty.resolved, f, false)
   def toJniCall(m: MExpr, f: String => String, needRef: Boolean): String = m.base match {
-    case p: MPrimitive => f(if (needRef) "Object" else IdentStyle.camelUpper(p.jName))
-    case MString => "(jstring)" + f("Object")
+    case p: MPrimitive => "(" + f(if (needRef) "Object" else IdentStyle.camelUpper(p.jName))
+    case MString => "reinterpret_cast<jstring>(" + f("Object")
```
artwyman: Generating the open-paren and close-paren at widely-separated places in the code seems confusing, and likely to miss some cases.  Why can't the close-paren be generated here so it's directly part of the cast expression?

artwyman (changes requested):
Thanks for updating.  Note that Djinni is on minimal maintenance, bugfixes-only, so don't expect a fast turnaround on reviews.  I'm willing to merge this in some form, though.  See my inline comments for some suggestsions, though.  The current implementation feels messy.

Note that if you want this merged you'll also need to sign the CLA here: https://opensource.dropbox.com/cla/individual/

src/source/JNIGenerator.scala
canatella: The problem lies in JNIGenerator.scala:288 where the expression to be casted is not entirely passed to the toJniCall method. I could add a function to generate the whole expression to be casted and then send that to toJniCall. It was not a problem before with c-style cast of course.

canatella: So after looking at it again, it does not really make sense to try to do that. Line 288 generates the first line of the call: something like

`auto jret = reinterpret_cast<jstring>(jniEnv->CallObjectMethod(`

Then on line 294 it uses that string length to generate properly aligned method call so it means the output of toJniCall should depends on the length of its output... Hoping I am clear here ;)

test-suite/generated-src/jni/NativeAssortedPrimitives.cpp
canatella: Due to the problem with the closing parentheses (see my following comments),  it would means that toJniCall should also return a value saying if there is a need to add the closing parentheses or not which starts to get things more complicated than they should be I think. If you have an idea on how to fix all those problems, I'll gladly give it a shot.

artwyman:
Thanks for your effort! This repo is no longer supported by Dropbox and about to be archived. I'm not going to try to merge this PR at the last minute, but I'll leave it open for reference by anyone who encounters the same problem.

PR #454, opened 2019-10-23
EXPERIMENTAL_AUTO_CPP_THREAD_ATTACH - add option to replace use of thread_local with pthread_create_key

tndhagedorn:
EXPERIMENTAL_AUTO_CPP_THREAD_ATTACH uses a thread_local object's destructor to detach a thread from the JVM
on exit - if it was previously attached due to a call across the JNI boudary.

The hooks to run a thread_local object's destructor have compatibility problems on some implementations
of Android L's libc, in call to `cxa_thread_atexit_impl`.

Add new define EXPERIMENTAL_AUTO_CPP_THREAD_ATTACH_NO_THREAD_LOCAL that does the same thing, but using
pthread_create_key/pthread_setspecific.  This has no dependency on `cxa_thread_atexit_impl`

### DIFF
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -21,12 +21,24 @@
 #include <cstdlib>
 #include <cstring>
 
+#ifdef EXPERIMENTAL_AUTO_CPP_THREAD_ATTACH_NO_THREAD_LOCAL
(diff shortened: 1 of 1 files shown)


tndhagedorn:
Apologies - please disregard.  Bad merge target in PR.

PR #456, opened 2019-11-21
Only set `-fobjc-arc` for valid languages on djinni_support_lib

remyjette:
`-fobjc-arc` is set as a PUBLIC compile option on djinni_support_lib,
meaning it also gets set for any targets that consume Djinni. However
the Swift compiler doesn't understand this option and will throw an
error if it is present. Any targets that contain Swift code and consume
Djinni, even if indirectly, will thus fail to build.

This change causes the `-fobjc-arc` option to only be set when the
language used for that compilation matches CXX, OBJC, or OBJCXX. CXX is
present because in CMake versions prior to 3.16 or if OBJC/OBJCXX is
not enabled, .m and .mm files are treated as CXX.


artwyman:
Thanks for your effort! This repo is no longer supported by Dropbox and about to be archived. Since CMAKE was community-submitted and not something I was set up to test, I can't validate if this is the right approach for all users. So I'm not going to try to merge it at the last minute, but I'll leave it open for reference by anyone who encounters the same problem.

PR #457, opened 2020-01-16
Update sbt-launch.jar to 1.3.7

remyjette:
Downloaded from https://dl.bintray.com/sbt/maven-releases/org/scala-sbt/sbt-launch/1.3.7/sbt-launch.jar

The existing version of sbt-launch.jar attempts to talk to Maven via HTTP
rather than HTTPS, which is no longer supported. Updating to the latest
sbt-launch.jar fixes the issue.


remyjette:
So turns out the issue was because I was running `sbt-launch.jar assembly` directly rather than the `sbt` script, as I was testing this on a Windows machine, but I wasn't previously passing the `java.security.poilicy` and `sbt.repository.config` args. Adding those fixed it, so I'll leave it up to you folks if you want to take this and update `sbt-launch.jar` anyway, or just close this PR without merging and leave things as-is.

